PR #6, opened 2019-02-18
Add doctype tag

jakul:
Fixes https://github.com/mrin9/RapiDoc/issues/2 and https://github.com/mrin9/RapiDoc/issues/3


mrin9:
Thanks for taking care of this one

PR #10, opened 2019-03-07
fix typo

birdayz:
i re-generated all dist files, is that correct?


mrin9:
Thanks for taking care of the issue. 

> Regenerating dist wont be a problem,

PR #16, opened 2019-03-15
Dereference all JSON references initially

mhailstone:
I tested an Open API 2.0 document that was very large and has many references and it failed. The `json-refs` module seemed to handle the size and position of the references better.
Signed-off-by: mhailstone <matthew.hailstone@gmail.com>

### DIFF
--- a/src/utils/parse-utils.js
+++ b/src/utils/parse-utils.js
@@ -1,5 +1,6 @@
 //import converter from 'swagger2openapi';
 import JsonSchemaRefParser from 'json-schema-ref-parser';
+import JsonRefs from 'json-refs';
 import converter from 'swagger2openapi';
 
 export default async function ProcessSpec(specUrl, resolveCircularRefs){
@@ -14,17 +15,19 @@ export default async function ProcessSpec(specUrl, resolveCircularRefs){
 
   let refParserOptions = {
     resolve: {
-      http: { 
-        withCredentials: false 
-      } 
+      http: {
+        withCredentials: false
+      }
     }
   };
   try {
     if (typeof specUrl==="string") {
-      convertedSpec = await converter.convertUrl(specUrl, convertOptions);
+      const resolvedSpec = await JsonRefs.resolveRefsAt(specUrl);
+      convertedSpec = await converter.convertObj(resolvedSpec.resolved, convertOptions);
     }
     else {
-      convertedSpec = await converter.convertObj(specUrl, convertOptions);
+      const resolvedSpec = await JsonRefs.resolveRefs(specUrl);
+      convertedSpec = await converter.convertObj(resolvedSpec.resolved, convertOptions);
     }
 
     if (resolveCircularRefs==='false'){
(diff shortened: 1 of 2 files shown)


mrin9:
Will it be possible for you to share the large spec? then it can serve as a test case, If possible, then you can put it under `./src/docs` folder for now. 
Also can you try out stripe's spec https://api.apis.guru/v2/specs/stripe.com/2019-02-19/swagger.json
with your changes ? there is this issue https://github.com/mrin9/RapiDoc/issues/14 which I closed, by providing a work-around, but if `json-refs` can handle this it will be really nice. 

> Thanks for your PR

mhailstone:
I stepped over the calls in Chrome's debugger and the call that makes the page wait is the `jsonParsedSpec = await parser.dereference(convertedSpec.openapi);` call on line 37 in parse-utils.js. If I set `resolveCircularRefs` to 'false' then the `jsonParsedSpec = await parser.bundle(convertedSpec.openapi);` on line 34 actually works. How is the `resolveCircularRefs` attribute calculated and passed in?

src/utils/parse-utils.js
```diff
@@ -14,17 +15,19 @@ export default async function ProcessSpec(specUrl, resolveCircularRefs){
...
+      const resolvedSpec = await JsonRefs.resolveRefsAt(specUrl, {resolveCirculars: resolveCircularRefs==='true'});
+      convertedSpec = await converter.convertObj(resolvedSpec.resolved, convertOptions);
     }
     else {
-      convertedSpec = await converter.convertObj(specUrl, convertOptions);
+      const resolvedSpec = await JsonRefs.resolveRefs(specUrl, {resolveCirculars: resolveCircularRefs==='true'});
+      convertedSpec = await converter.convertObj(resolvedSpec.resolved, convertOptions);
     }
 
     if (resolveCircularRefs==='false'){
```
mrin9: You dont need this` if` and the `else` block at all, if you are dereferencing using `json-refs`.  `parser` in this code  is an instance of `json-schema-ref-parser` and `parser.dereference` is same as `JsonRefs.resolveRefs`

Now with `json-schema-ref-parser`, for certain It fails to dereference properly and leads to  non-responsive browser, therefore as a workaround there is an attribute provided at the `<rapi-doc>` element `resolve-circular-refs`
its default value is `'true'` but in case some spec is hanging up, then user of the `<rapi-doc>` element can decide not to resolve the refs, in that case we use json-schema-ref-parser's `bundle()` method which wont resolve the refs. This will at least show up the spec in the UI but wherever it has `$ref` it will try to lazily resolve

So remove this block for now, `resolve-circular-refs` attribute too will not have any purpose, if you want to clean up that code then you need to start from `rapidoc.js` and then it is passed down to the lower components, if you want to skip it, thats ok, I will do it later.

Now `parser` object being redundant should be removed, but it is passed to lower components for lazy resolving of `$refs`, refer to line `188` of this file where it is passed this too needs to be cleaned up, else we wont be able to get rid of  `json-schema-ref-parser`.

Goal should be to use either of the parser not both and keep our dependencies and bundle size to minimum and.

mrin9: I am accepting the PR for now, and let me slowly remove things that are not needed.
Thanks for your contribution, 

`json-refs` is indeed better than `json-schema-ref-parser` when it comes to handling large specs. but it increases the bundle size from `98 kb` to `140 kb` (gzipped) but it is a better tradeoff

PR #19, opened 2019-03-20
Added URL encode to path params, query params and the API key as a qu…

mjbrisebois:
Thought I would take a shot at adding the fix myself.  However, I was not able to test the code.


mrin9:
Thanks for your contribution

PR #38, opened 2019-07-01
Show server descriptions if they exist

giodamelio:
First off I would like to say I really like this doc renderer. It is both pretty and customizable!

I was evaluating it earlier and I noticed that the server selector doesn't show the descriptions if they exist. So this just adds them in right after the url. Not sure if that is the best way to show them, but I figured I'd give it a try.


mrin9:
thanks for the PR, this never crossed our mind.

PR #50, opened 2019-08-24
Fixes inputs to support multi security schemes

kongdigital:
The current UI only supports 1 value in the security definitions. Multiple values are not set and are not used when sending a request.


mrin9:
I am not sure when you say multi security schemes are not supported. 
have you checked [example8](https://mrin9.github.io/RapiDoc/example8.html)

which has multiple security schemes (http and https) is that not working for you ?

kongdigital:
@mrin9 Let me clarify in the example you provided you can only set 1 authorization. Setting a value in 2 or more fields only uses the last value. The Open API spec supports using multiple values as does the default swagger ui.

And the use case is requiring both an `x-api-key` header and a `Authorization` header in the request.

mrin9:
I see what you mean now. 
The way you are doing it, is by introducing another custom-property to the web-component.

I appreciate the effort on the PR. but the way, I like to drive it is by not introducing any custom-property, instead use the current UI in a way where setting multiple security scheme will not remove the previous value, plus there should be an option to clear a specific security. 
That way, people can dynamically generate any combination of security schemes needed by the api server.

I would be very much interested if you can help me on this and implement it in the way, it is described above

mrin9:
closing the PR in favor of commit 444a6a9

PR #55, opened 2019-09-24
Make it possible to hide API Server selection

FlaxHaxx:
This adds the new  the property `showServerSelect` to the `rapi-doc` component. This will hide the API Server select section when it is set to `false` and the property `serverUrl` will be used if it is set, otherwise the first url in the Swagger/OpenApi spec will be used.


mrin9:
I am trying to understand the motivation behind it 
- is it only to control the UI rendering 
- is it to restrict an user to select an API server, when multiple server exist in the spec ? in that case how would you restrict it to the 2nd, 3rd or 4th (not first) from the API server list 
- how are you informing the user which API server he/she is connecting to ?

FlaxHaxx:
Thanks for looking at my suggestion.

It is only for UI rendering. The use case I have is a developer portal with a full API reference where I want to use the existing OpenAPI specification files without modifying them. The API is built using microservices where each microservice has it's own OpenAPI yaml file. You can see my use case at https://www.northbricks.io/apis/.

Maybe it's better to require the use of both the `server-url` attribute and this new attribute to hide the server select section. This way we are sure the right URL is used when connecting to the API with the *Try* button.

The user can see which API server he/she is connecting to by expanding the operation. In my case I can also show it by writing it somewhere else on the website, outside the RapiDoc component.

mrin9:
ok, I think for certain UI scenarios this must be a needed feature like you showed me one of yours. 
I have couple of comments 
- lets name the attribute as `allow-server-selection` with a default value of `true` and this will be effective only if `allow-try` is  `true`
- please update the documentation where we list all the attributes and a short description about it at `./src/index.html` and `./src/welcome.html` both files are identical, one is used for testing internally
- I hate to say this but we have diverged a lot since you created this PR and have applied certain code-style rules. Can you submit another PR that confirms to new standards ?

FlaxHaxx:
Created a new pull request from your comments. Please see #56

PR #65, opened 2019-10-22
Set default selection to first server with allowServerSelection

### LINKED ISSUE #64: selectedServer undefined after 6.0.0 release
After the release of v6.0.0 we have problems with our hosted RapiDoc based documentation, and it seems to be related to the change in:
https://github.com/mrin9/RapiDoc/commit/5da2c44b51bd357079b615e02b52b63693c9c164

From:
https://github.com/mrin9/RapiDoc/blob/8057012c7b1a45ddfe7b0e50ad46f25be33b8be4/src/rapidoc.js#L642-L648
To:
https://github.com/mrin9/RapiDoc/blob/5da2c44b51bd357079b615e02b52b63693c9c164/src/rapidoc.js#L720-L724

Combined with the fact that the initial `setTimeout()` call to `onApiServerChange()` is not passing the value of the first server in:
https://github.com/mrin9/RapiDoc/blob/8057012c7b1a45ddfe7b0e50ad46f25be33b8be4/src/rapidoc.js#L705-L707

So this means the initial value of the `selectedServer` is `undefined`

peterbroadhurst:
This resolves #64 in my environment

With the default of `allow-server-selection` enabled, it is still necessary to set a default value for `selectedServer` so that the first entry in the radio-button list is selected.

There is also no need to call `this.onApiServerChange()` as it's now a no-op when there's no `e` passed in.

### DIFF
--- a/src/rapidoc.js
+++ b/src/rapidoc.js
@@ -780,10 +780,12 @@ export default class RapiDoc extends LitElement {
 
   afterSpecParsedAndValidated(spec) {
     this.resolvedSpec = spec;
-    if (this.serverUrl) {
-      this.selectedServer = this.serverUrl;
-    } else if (this.resolvedSpec && this.resolvedSpec.servers && this.resolvedSpec.servers.length > 0) {
-      this.selectedServer = this.resolvedSpec.servers[0].url;
+    if (!this.selectedServer) {
+      if (this.serverUrl) {
+        this.selectedServer = this.serverUrl;
+      } else if (this.resolvedSpec && this.resolvedSpec.servers && this.resolvedSpec.servers.length > 0) {
+        this.selectedServer = this.resolvedSpec.servers[0].url;
+      }
     }
     if (!this.apiListStyle) {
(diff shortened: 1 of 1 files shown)


mrin9:
yep `this.onApiServerChange()` is obsolete  thanks for taking that out

there is this attribute `default-api-server` (yet to be documented). it allows you to provide the preferred api-server https://github.com/mrin9/RapiDoc/blob/0ff9b2fc91faae02ea2221c7b190e4fa508b2b19/src/rapidoc.js#L60
If provided and it matches any os the available servers, it should trump the logic of selecting the first server in the list, can you make that change too ?

peterbroadhurst:
Thanks @mrin9 - I've updated the PR to honour `default-api-server` if set.

mrin9:
if `default-api-server` is provided but is wrong or do not match any from the available list of servers then the radio buttons will be left unchecked.
I am merging it now and take care of this  gap later

peterbroadhurst:
Big thanks @mrin9 for the fast release 💪 🚀

PR #66, opened 2019-10-23
Two crashes fixed

vovayartsev:
I'm trying out RapiDoc in an Elixir / Phoenix app and https://github.com/open-api-spex/open_api_spex library as OpenApi spec generator.   Somehow I managed to create `swagger.json` which was crashing RapiDoc (my bad).

I thought it's generally a good idea to be more "liberal on the input" - so I fixed the crashes in the most strait-forward way I could.  I'll be happy to apply your suggestions to make it a "proper" fix. 

Thanks!

### DIFF
--- a/src/rapidoc.js
+++ b/src/rapidoc.js
@@ -15,6 +15,8 @@ import '@/components/end-points-expanded';
 import '@/components/path-and-methods';
 import '@/components/security-schemes';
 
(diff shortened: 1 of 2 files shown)


mrin9:
Can you provide me a test spec for me to test against

mrin9:
closing due to inactivity

PR #90, opened 2019-11-15
Remove char

karlkilgi:
Seems like a small unwanted character was left in here, removed it.

Produced Í after every title.


mrin9:
yep we noticed it, thanks for taking care of it. We fixed it internally so I will have to reject this PR. I am sorry.

PR #124, opened 2020-02-03
Fix regression in Swagger 2.0 support in 7.2.0 - missing path in server URLs

peterbroadhurst:
Fix for https://github.com/mrin9/RapiDoc/issues/123


mrin9:
Thanks much - for investigation and the PR

PR #151, opened 2020-03-16
Feature/oauth

fbacker:
Rework on authentication. 
I've created example html/spec file (authentication.html/json) to show the different scenarios and gui. Tried OAuth2 flows with Identity Server setup (tried swashbuckle gui and then this and both worked).

1. More information in authentication list
2. add OAuth2 clientCredentials and authorizationCode support
3. Request information more info about required security
4. Request each authentication is seperated to own calls

**note:** if more are interrested to have more auth types maybe a look into swaggers own auth code should solve some things more generic. But I think this should work as a start.

Issues that talks about oauth
- https://github.com/mrin9/RapiDoc/issues/142
- https://github.com/mrin9/RapiDoc/issues/54

image


mrin9:
Thanks for the PR, Supporting OAuth has been a long pending feature, but there is no suitable workflow that we could come up with in the context of Custom-Element 

For instance as a part of this PR, you created a file `docs/oauth2-redirect.html` which helps in token exchange, but this file will not be a part of the build output and therefore will break the flow

The build script generates a single `RapiDoc.js` file which is a pure java script Custom-Element no `css`, `fonts`, `images` or anything else is included with it

If this would have been a WebApp like [MrinDoc](https://mrin9.github.io/OpenAPI-Viewer/) we could have included `docs/oauth2-redirect.html` and can place it on the WebServer. but thats not the case with Custom Elements.

let me know your thoughts

fbacker:
Could it be that you mention the api docs and/or github readme. 

That if ```authorizationCode``` is used the ```oauth2-redirect.html``` file is needed to finalize the authentication. It should be placed in the path root or use the ```redirectUrl``` parameter in specification file. 

Then a direct link to the file in github where the developer can download the needed oauth html file.

If we get the html file in dist folder the file could be accessed thru ```https://unpkg.com/rapidoc/dist/oauth2-redirect.html``` so quick access for developer would just to add that as the ```redirectUrl```. 

Think it's inside the scope with OAuth to be a little more configuration. You would need a oauth server, setup client credentials, redirect urls and so on. 

Can try to help write some kind of documentation help if you want?

mrin9:
The way I am envisioning it is to create another custom-element something like `<oauth-receiver>`  which will contain the functionality of `oauth2-redirect.html`

and some documentation which will state to create another `.html` file and include `<oauth-receiver>` in it.

fbacker:
I've started a new branch from the latest master and trying to adjust the authentication to match. Think what I'm doing now will work better and easier to implement all different kind of oauth types in a later stage.

So this branch we could throw out and I can create a new PR when I'm ready.

fbacker:
I've a much better pull request for this now. Will send a new pull request soon. Needs just a little more testing for many scenarios.

Issue #21, opened 2019-04-23
Multiple API Key headers


Paxa:
Now it will use only first value

example
```json
  "security": [
    {
      "ClientId": [],
      "PassKey": []
    }
  ],
  "components": {
    "securitySchemes": {
      "ClientId": {
        "type": "apiKey",
        "in": "header",
        "name": "Client-Id"
      },
      "PassKey": {
        "type": "apiKey",
        "in": "header",
        "name": "Pass-Key"
      }
    },
```

mrin9:
This should be fixed in 5.0.3

PR #154, opened 2020-03-18
Feature/oauth2

fbacker:
A rebuild from https://github.com/mrin9/RapiDoc/pull/151 PR with better and simpler code usage.

Tested oauth on real sever
- clientCredentials
- authorizationCode

Prepared but not tested (don't have oauth configuration server). I've placed code and @todo tags pointing to where to look and test.
- implicit
- application
- password
- using pkce

Features
- Support for multiple flows in same security scheme
- operation lists required scopes
- scope selection when requesting tokens
- using new oauth-receiver html file
- turn on/off feature thru 'api'

API Features from html **(default value)**

- show-authentication-redirect-info **(true)**
    Show extra auth setup information when redirect url exists and/or authorizationCode is used 
      
- show-authentication-operation-info **(true)**
    Show required scheme security id and scopes for operation

- allow-authentication-multiple-call **(true)**
    If multiple auth keys exists, allow to use them all in one call

- allow-authentication-single-call **(true)**
   If multiple auth keys exists, allow calling each one specifically


fbacker:
@mrin9 did you have a change to look at this, so you don't do a lot of work on the old PR.

fbacker:
Example on complexed oauth scopes
image

And making a request
image

fbacker:
Something that needs to be fixed is when a request is made and if there's an error, it's intuitive to need to check the developer console.

mrin9:
oops, I missed the update. Let me review it this weekend, I already there are some conflict.

fbacker:
Saw that you had made a lot of commits. Tried to merge to latest master. Hope that this solvs our oauth needs. As described above, solves the standard and some code exists for implementing the other types. I don't have the possibility to try it out. Looked how swagger ui solves their and tried to mimick it matching your project structure.

package.json
```diff
@@ -28,11 +28,15 @@
...
+    "@braintree/sanitize-url": "^4.0.0",
+    "crypto": "^1.0.1",
+    "jshint": "^2.11.0",
     "lit-element": "2.2.1",
     "lit-html": "1.0.0",
     "marked": "^0.8.0",
     "swagger-client": "^3.10.0",
-    "swagger2openapi": "^5.3.4"
+    "swagger2openapi": "^5.3.4",
+    "url-parse": "^1.4.7"
```
mrin9: I would like to keep our bundle size to minimum as we do not do any code splitting.  I dont think any of the packages added here is needed for our use-case

src/templates/security-scheme-template.js
```diff
@@ -36,99 +41,168 @@ function onClearAllApiKeys() {
...
+    });
+  }
+
+  window.Oauth2UIRedirect = {
+    flow,
+    state,
+    redirectUrl,
+    callback,
+    authConfigs,
+    error: (err) => console.error(err),
```
mrin9: `Oauth2UIRedirect` a custom attribute is created for window-to-window communication, Instead I would prefer the standard approach [`window.postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) and [`messageEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent) which currently being used

src/components/api-request.js
```diff
@@ -49,6 +50,9 @@ export default class ApiRequest extends LitElement {
       activeSchemaTab: { type: String, attribute: 'active-schema-tab' },
       schemaExpandLevel: { type: Number, attribute: 'schema-expand-level' },
       schemaDescriptionExpanded: { type: String, attribute: 'schema-description-expanded' },
+      showAuthenticationOperationInfo: { type: String, attribute: 'show-authentication-operation-info' },
+      allowAuthenticationMultipleCall: { type: String, attribute: 'allow-authentication-multiple-call' },
+      allowAuthenticationSingleCall: { type: String, attribute: 'allow-authentication-single-call' },
```
mrin9: currently RapiDoc supports providing multiple or single API keys through the UI. Users can add or remove any number of keys using the UI. There wont be any need for any attributes.

It will also create confusion if we set `allow-authentication-multiple-call` and `allow-authentication-single-call` both to true

Moreover it will create a breaking change for #21

PR #156, opened 2020-03-19
Added support for live search

ethanabowen:
RapiDoc search can now look through your API as you type your search value by adding allowLiveSearch="true"

### DIFF
--- a/src/rapidoc.js
+++ b/src/rapidoc.js
@@ -65,7 +65,7 @@ export default class RapiDoc extends LitElement {
       allowSpecUrlLoad: { type: String, attribute: 'allow-spec-url-load' },
       allowSpecFileLoad: { type: String, attribute: 'allow-spec-file-load' },
       allowSearch: { type: String, attribute: 'allow-search' },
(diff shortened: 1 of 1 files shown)


ethanabowen:
This probably should have just been a branch merge and not a coming from fork.  If you'd prefer the branch approach let me know and I'll do that.

ethanabowen:
This is on my radar and I am expecting to get to this quick change within the week.  Thanks.

mrin9:
There is no hurry, take your time. Just an heads up there is a pending PR #171  that I am reviewing. It adds the functionality of collapsible tags for `render-style = view`, if a tag is collapsed and someone is searching for a path inside the collapsed tag, it will automatically expand itself. 

Making a a live-search may expand many tags if it is fired for every character we type, I think it will be a good idea if we add some debouncing (delayed) event
You dont need to  take care of it in this PR if you are short on time, but just keeping you infomed

ethanabowen:
I'm closing this since so much time has pasted and I've run out of time to implement this right now.

Issue #161, opened 2020-03-22
Unexpected Code Block Indentation


BinaryWizard904:
Under current styling, the beginning and end of text within multi-line code elements are indented as seen here:
![RapiDoc Code Indentation](https://user-images.githubusercontent.com/62070150/77253820-f8fd5680-6c2a-11ea-8529-034bc5ffa0d6.png)
This is due to the horizontal padding found here: https://github.com/mrin9/RapiDoc/blob/759a18340e49ad7eddaa24953608bdf136625e27/src/styles/font-styles.js#L97
As code is expected to be reproduced verbatim, this styling implies to users that there is a preceding or trailing space in the code.  I propose we shrink the padding to eliminate this misperception.

mrin9:
Yep, there are other issues too. 
- Multiline code block shouldn't be red color  thats for inline mono-spaced font
- background color should be spanning the whole block instead of limiting it to text-span

Something like below

![image](https://user-images.githubusercontent.com/5056899/77262779-adab6e00-6c53-11ea-8212-15a130aa0218.png)

The PR you submitted wont cover all the cases plus it will disturb the other cases
will get it fixed soon in a separate commit

PR #162, opened 2020-03-22
Remove Horizontal Padding of Code Blocks

BinaryWizard904:
As per #161, the previous padding caused the beginning and end of a multi-line code block to be indented.  This incorrectly implied to users that a preceding or trailing space was present in-code.  This change reduces the size of the padding to eliminate this misperception.


mrin9:
Thanks for the PR, but this will disturb the styling of single line `mono-spaced` block which needs horizontal padding

We need target only the multiline code blocks. I am going to reject this PR and make some styling changes as it need to address certain other issues apart from padding

PR #163, opened 2020-03-23
allow multiple oauth2 types

fbacker:
Lets try this again then.

built on top of 2 PR https://github.com/mrin9/RapiDoc/pull/154 and https://github.com/mrin9/RapiDoc/pull/151

I've started a new branch from master and moved my old PR code into this. Adjusting after feedback from previous PRs.

I don't understand how to easy populate subclasses with settings from html, that can be a little messy.

Solves OAuth2 security with multiple different schemas.

Tested oauth on real sever

- clientCredentials
- authorizationCode

Prepared but not tested (don't have oauth configuration server). I've placed code and @todo tags pointing to where to look and test.

- implicit
- application
- password
- using pkce (needs to create hashes e.g. crypto library)

Features

- Support for multiple flows in same security scheme
- operation lists required scopes
- scope selection when requesting tokens
- using new oauth-receiver html file
- turn on/off feature thru 'api'

```
hide redirect information (only needed for configure developer, not everyone else when done)
    hide-authentication-redirect-info="true" 

show needed security requirements on each operation (scopes)
    show-operation-requirements="true" 

seperate calls based on security. When using different security clients/scopes on calls. make it possible to call with correct scheme
    allow-authentication-separated-calls
```

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -44,11 +44,14 @@ export default class ApiRequest extends LitElement {
       responseStatus: { type: String, attribute: false },
       responseUrl: { type: String, attribute: false },
       allowTry: { type: String, attribute: 'allow-try' },
+      allowAuthenticationSeperatedCalls: { type: String, attribute: 'allow-authentication-separated-calls' },
       renderStyle: { type: String, attribute: 'render-style' },
       schemaStyle: { type: String, attribute: 'schema-style' },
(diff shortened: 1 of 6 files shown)


mrin9:
@fbacker , I am closing this as release `7.5.0` is out has the OAuth support. In case you still having issues to make it work on your end, please feel free to report them.

Thanks a lot for the Initial PR, which I used as a base to implement this feature

PR #172, opened 2020-04-02
allow duplicate paths on tags

fbacker:
A feature to allow the usage of multiple tags. For every tag the operation will be displayed (default in swagger ui).

By setting ```allow-duplicated-paths-by-tag="true"``` an duplicate of the operation will be created for each tag. Can be tested on ```multiple-tags-from-paths.html```.


src/utils/spec-parser.js
```diff
@@ -177,83 +177,93 @@ function groupByTags(openApiSpec, sortTags = false, sortEndpointsBy) {
...
-                return commonParam;
-              }
-            }).concat(fullPath.parameters);
+          tagObj = tags.find((v) => v.name === tag);
+          if (!tagObj) {
+            tagObj = {
+              show: true,
+              name: tag,
+              description: tagDescr ? tagDescr.description : '',
+              expanded: tagDescr ? tagDescr.expanded : true,
```
mrin9, line 214: I dont get this logic ?

fbacker: Sorry about that row.
I've multiple PR for this project. Expanded is related to the collapsable. Have another branch where all my PR are merged so I can use them. Expanded can be removed from this branch.

fbacker:
@mrin9 this is ready for testing. updated with latest master. 
I've put it in list.html for example and added api docs.

PR #181, opened 2020-04-08
Added support for x-code-samples and tab panels inside endpoint descriptions

dcarr178:


mrin9:
I haven't looked into in much detail, but from quick browse it seems like you are creating a DSL in a html style comment block and then rendering it in RapiDoc.

if thats the case, then I will request a change. This approach locks down the spec with the spec-viewer. The same spec will render ugly in other viewers such as swagger-ui because html comment blocks do appear in the markdown output 

We have recently updated RapiDoc with syntax highlighting, our next task on this line is to support `x-code-sample` which is a popular vendor extension. 
And the way we would like to achieve is by providing it in the spec using `x-code-sample` 

you can look at the sample spec here https://github.com/mrin9/RapiDoc/blob/master/docs/specs/petstore_extended.yaml#L145-L173

can you modify your code to support `x-code-sample`  instead of putting it in markdown ?

also it will be quicker for me to get a UX approval if you put a screen-shot :)

dcarr178:
> you are creating a DSL in a html style comment block and then rendering it in RapiDoc

This is true

> The same spec will render ugly in other viewers such as swagger-ui because html comment blocks do appear in the markdown output

I have tested it using swagger-ui and verified that html comments inside endpoint descriptions do not appear. However, swagger-ui support for standard markdown inside endpoint descriptions is horribly limited. They don't support even basic code blocks.

> can you modify your code to support x-code-sample instead of putting it in markdown ?

Yes - this PR supports both and gives users both options.

/examples/x-code-samples.html
image

/examples/tab-panel.html
image

mrin9:
This is an awesome feature that you added, got a lot of good feedback.

There are certain aspect that needs some change and minor implementation issues that we found in our testing. 

Below are some top things that was discussed
- If code samples needs to be provided through markdown then should be done in the standard markdown way which is 

  ```
    ```java
      some code...
    ``
  ```
   wont support a DSL that encourages a vendor lock-in. (the spec may not work the same in other viewers)

- A TabPanel custom element  was created, this needs to be implemented  as Template instead. 
 >  _We are trying to move away from creating too many custom elements, because custom element creates shadow-dom (shadow boundaries) this stops styles and dom access methods to cross these boundaries. We also at some point like to do a server side rendering, but rendering web-components on server is non-trivial_

- There was some minor glitches with the tab-component, for instance clicking on the empty section of tab-strip was making every thing disappear 


However, I will be merging your PR and make the necessary changes.

Thanks for helping us with this awesome feature !!!

PR #184, opened 2020-04-14
Client Cred flow following https://tools.ietf.org/html/rfc6749#section-4.4.2 

pavlovmike:
The recommended way for Client Cred flow authentication by base Oauth RFC


mrin9:
Thanks for testing and fixing it per the standards !!!

Issue #179, opened 2020-04-07
Would it be possible to reload the spec after authentication status changes?


bytedreamer:
The goal is to only show the API endpoints that are authorized. A call to retrieve the spec should include the authorization header so that endpoints could be filtered.

mrin9:
we have a similar task on our todo, but thats to annotate (may be disable) the try button. We try not to do any validations, to keep the API testable, for instance
- if you defined some constrain on a schema field like max-length, RapiDoc wont stop if you provide more than the max-length, this helps user to test how the API handles the negative cases ,
- same will be with security, would like to keep the end-point open to allow testing APIs that are invoked without security
 
Is there a use-case that you are after ?

bytedreamer:
Certain calls are restricted based on the user's permissions. We want the user to be able to view only the calls for which they have access.

The workflow for an user accessing the API Documentation:

1. Page shows no calls, only the OAuth2 section
2. User logs in using our OAuth2 login page
3. Successful login will only show the calls that the user has permission

One way to accomplish this is to dynamically create JSON spec file based on user permissions. For this to work, the JSON spec file needs to be reloaded anytime the login status changes. The JSON retrieval also needs to include the OAuth2 bearer token after login.

mrin9:
I dont think at present we will be taking that route. but keep an eye on this space, we plan to programmatically allow hide certain paths. once we provide that feature you can do it at your end. I will provide an event when a auth-key is applied, that will help you do what you exactly want.

PR #188, opened 2020-04-16
add token to spec load

rinella:
I modified the loadSpec to send a token with the request so the swagger json can be filtered based on what a user has permission to in the api.  i was wondering what the use case is for multiple tokens and if this is something you would be interested in for the mainline.

thanks.

-Eric

### DIFF
--- a/src/rapidoc.js
+++ b/src/rapidoc.js
@@ -841,8 +841,15 @@ export default class RapiDoc extends LitElement {
     this.matchPaths = '';
   }
 
+  async refreshSpecWithToken(apiKeyId, msgEvent) {
+    await this.loadSpec(this.specUrl, msgEvent.data.access_token);
+    const securityObj = this.resolvedSpec.securitySchemes.find((v) => (v.apiKeyId === apiKeyId));
+    securityObj.finalKeyValue = `Bearer ${msgEvent.data.access_token}`;
+    this.requestUpdate();
+  }
+
   // Public Method
-  async loadSpec(specUrl) {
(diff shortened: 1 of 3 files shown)


rinella:
reference issue #179

mrin9:
Thanks for the PR, as I mentioned in #179 we have a different opinion on how we are going to approach operations with security.

We are unlikely to take this PR, 

but I still did a quick code-gaze, it looks all fine, However, I would have taken a reactive approach
 
For a small to mid sized spec, this approach may work fine, but for larger specs it may not cut the deal. From your code it seems like it will trigger a reload, parse of the whole spec, and after that will re-render of the whole app, 

I dont think a reload is necessary at all, we already have a parsed spec with us and the endpoints can be hidden/shown in a reactive manner based on which API keys are applied and which Endpoints can be used with the applied API Keys.

This is how the search functionality works, the Endpoints are hidden/shown based on the search text and its presence in the endpoint path

rinella:
Thanks for checking it out.  Totally understand where you are coming from on the performance aspects.  

Our requirements are a little more security slanted.  We do not expose any endpoints unless you are authenticated.  The initial spec is empty.  Upon spec reload with the token in the request, the endpoints the user has access to are exposed.

mrin9:
If you plan to update your fork, to hide/show endpoints reactively based on if an API key is applied, please consider a PR here. We do have plans to implement similar functionality to filter out operations based on various aspects like Security, Tags. We can use your PR and extend it to support some of our workflows

PR #190, opened 2020-04-22
Fix trailing semicolon in 'Content-Type' header

trebler:
Trailing semicolon in `Content-Type` header is invalid. 

`express.js` fails to parse request body sent with such a header.


trebler:
@mrin9

mrin9:
Thanks much for spotting and fixing it !!!

PR #191, opened 2020-04-23
fix trailing spaces after line breaks in curl command

trebler:
Having spaces after line breaks (`\`) makes `curl` (or any shell command for that matter) ignore following lines, and hence `curl` example is not immediately usable when copying from `RapiDoc`.

@mrin9


mrin9:
awesome !!! thanks for the PR

PR #199, opened 2020-04-29
Fixed expanding of operation for "view" mode

Xenar0:
During navigating either via URL hash changes or via "scrollTo" public method, there is a bug when the view for such kind of operations isn't expanded:
- post-/pet/{petId}/uploadImage
- get-/pet/findByStatus

it works only for "one-section" paths, like
- put-/pet


mrin9:
thanks for identifying and fixing this issue !!!

PR #202, opened 2020-04-30
Display image

bytedreamer:


meztez:
Could it be generalized to all images format instead of only png?

mrin9:
First, thanks much for taking time and trying to address certain concerns with this library. We do appreciate your contribution and help us make it a better product.

However, I will have to discard this PR, for
- one we do not want to add token to load a spec
- two, it is a single PR trying to address multiple concerns, this do no not help much in getting through our internal approval process

for future PRs, please create an issue first and then submit a PR against it

mrin9:
there is a separate issue #251  exist for displaying images, will address it to support multiple other formats not just `png`
it should support `image`, `video`, `audio`, `zip` etc

We do support receiving binary content, however for that the response `content-type` in the spec should be `binary/octet-stream` or `application/octet-stream`
we will now add more binary format to that list

PR #222, opened 2020-05-27
Remove implicit scroll during on intersect with observable items

ilya-edrets:
Hi,

Our team started to use RapiDoc for our developer portal and we are really excited with getting results.
We plan to embed RapiDoc in the read mode as a page of our portal.

But, I run into a problem. If I wrap `<rapi-doc>` with any html tag, then scroll doesn't work. For example
``` html
<html>
  <body>
    <script src="js/rapidoc-min.js"></script>
    <div>
      <rapi-doc spec-url="openapi.yaml" render-style="read"></rapi-doc>
    </div>
  </body>
</html>
```
I've reproduced this problem in Chrome 83 and Firefox 76.

After some research I find out that problem in `onIntersect` method which calls `scrollIntoView` on intersected element.
It looks redundant, because if `onIntersect` is called, it means that the element is already visible and it's don't require scrolling to it.
I just remove `scrollIntoView` and the scroll starts to work correctly.

Can you take a look at my PR, please?

Thanks.

### DIFF
--- a/src/rapidoc.js
+++ b/src/rapidoc.js
@@ -416,7 +416,6 @@ export default class RapiDoc extends LitElement {
         // Add active class in the new element
         if (newNavEl) {
           window.history.replaceState(null, null, `${window.location.href.split('#')[0]}#${entry.target.id}`);
-          newNavEl.scrollIntoView({ behavior: 'auto', block: 'center' });
           newNavEl.classList.add('active');
         }
         // Remove active class from previous element


mrin9:
Thanks the PR, 
however the PR will break a functionality, In read mode when you scroll the main-content, 
the `nav- bar` must automatically focus the selected menu-item based on the scroll position of main content and get the highlighted menu-item into the view.

However, lets try to fix the issue at hand where you mentioned that scroll wont work properly in read mode.

There are two ways to address this 
- Define a height of `<rapi-doc>` component, you can directly specify it using the style attribute or it can be inherited from a parent component.
have a look at it here https://codesandbox.io/s/upbeat-sea-4tfo7?file=/src/App.js

```html
<html>
  <body>
    <script src="js/rapidoc-min.js"></script>
    <div>
      <rapi-doc 
        spec-url = "openapi.yaml" 
        render-style = "read"
        style = "height:100vh"   <!--  Specify  height -->
      >
      </rapi-doc>
    </div>
  </body>
</html>
```


- we will be introducing a new `render-style` - `focused` this looks like a `render-style='read'` except that it shows only one operation at a time and do not activate scroll-observer. the code is already baked in, you need to just build it on your end and use `render-style='focused'`

Let mw know if you still facing any issue on this

ilya-edrets:
Thanks a lot for your advice. It fix my problem.

PR #225, opened 2020-05-29
Change expand/collapse text to singleline/multiline in schema table

### LINKED ISSUE #219: schemaDescriptionExpanded button in schema table cause misunderstanding
Regardless of value of `schema-description-expanded`, there is `toolbar-item` button (named 
Expand/Collapse Details) in schema table. See the below examples.

image

image

After clicking the button, it has no effect to the schema table if all description are one line.
I think It make user misunderstand that the button is used to expand/collapse schema level and it is broken now.

Suggestions:
1. Hide the button if `schema-description-expanded` is `true`
2. Modify the button text

winonecheng:
Fix #219
Following commit [0cde5d0](https://github.com/mrin9/RapiDoc/commit/0cde5d0e841775295b6b99b6c8e1a280b8b9e63d)


mrin9:
Thanks for the PR !!!

PR #247, opened 2020-06-25
fix Content-Type header to be according to IANA and RFC

Uzlopak:
We use expressjs and the body parser is implemented in a strict way. RapiDoc is always adding charset: utf-8 even though this is non-standard and not according to the IANA-Definition. 

This PR changes the api-request.js file. If the Content-Type is application/json and uses only that as for the header. If it is not that, it will use the old method. This implementation uses a switch to extend the functionality if there are other use cases. It also provides a comment,which is a link to the official IANA-page, which should enforce the clarity of the line.

See also:
https://github.com/request/request/issues/383

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -968,8 +968,17 @@ export default class ApiRequest extends LitElement {
       // Common for all request-body
       if (!requestBodyType.includes('form-data')) {
         // For multipart/form-data dont set the content-type to allow creation of browser generated part boundaries
-        fetchOptions.headers['Content-Type'] = `${requestBodyType}; charset=utf-8`;
+        switch (requestBodyType) {
+          // https://www.iana.org/assignments/media-types/application/json
+          case "application/json":
+            fetchOptions.headers['Content-Type'] = "application/json";
+            break;
+          default:
+            fetchOptions.headers['Content-Type'] = `${requestBodyType}; charset=utf-8`;
+            break;
+        }
       }
+      
       curlHeaders += ` -H "Content-Type: ${requestBodyType}" \\\n`;
     }
 


ckeboss:
Funny enough, I also just created this issue: https://github.com/mrin9/RapiDoc/issues/248

Although, I don't think your solution goes far enough. If the OpenAPI spec you load specifies a content type, I believe RapiDoc should use that content type exactly. If the content type in the OpenAPI spec is `application/json; some-param=1` or `application/json; charset=utf-8=1`, it should simply use that, and not try to append to it.

Uzlopak:
Well I didnt want to break anyones code, as application/xml can have charset and makes actually sense as I guess the only way that the charset of the payload is not utf8 when as such the corresponding charset tag is defined in the HTML header.

mrin9:
I am not sure if I get everything that @Uzlopak  wants to convey,

but if we are talking about a situation where an `XML` payload can contain a `charset` inside it, then that wont be following the standard [RFC 6838](https://tools.ietf.org/html/rfc6838#section-4.2.1)

```
 As specified in [RFC6657], a "charset" parameter SHOULD NOT be
   specified when charset information is transported inside the payload
   (e.g., as in "text/xml").
```

So I would request to change the PR to just remove the extra `charset` param and use it as is without any modification. like @ckeboss suggested

src/components/api-request.js
```diff
@@ -968,8 +968,17 @@ export default class ApiRequest extends LitElement {
...
+          // https://www.iana.org/assignments/media-types/application/json
+          case "application/json":
+            fetchOptions.headers['Content-Type'] = "application/json";
+            break;
+          default:
+            fetchOptions.headers['Content-Type'] = `${requestBodyType}; charset=utf-8`;
+            break;
+        }
       }
+
```
mrin9, line 981: just removing the extra `charset` option would be enough, do not want to add `charset=utf8` for any other content type including `xml`

PR #250, opened 2020-06-26
Add support for enums in server veriable object

siavashs:
Add support for enums in server variable object.
Variables with an `enum` are rendered as `select` with the `default` option set to `selected`.
Variables without an `enum` are rendered as `input` (like before).
The examples are updated to show the above cases.

Reference: https://swagger.io/specification/#server-variable-object

### DIFF
--- a/docs/specs/server.yaml
+++ b/docs/specs/server.yaml
@@ -4,19 +4,23 @@ info:
   title: Server Variables
   description: >
     Various ways to define API servers in the spec including variables
-    
+
     ```yaml
       servers:
         - url: 'http://example.com/api'
           description: Test server (just for example)
         - url: '/api'
-        - url: '{protocol}://dev.com/api'
+        - url: '{protocol}://dev.com/api/{version}'
           variables:
             protocol:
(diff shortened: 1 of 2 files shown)


siavashs:
@mrin9 I confirm:
image

siavashs:
BTW, I think we also need an example with a variable which only has the `default` and no `enum`, that should create a text `input` instead.
I tested that case with my own `swagger.json` copy but I can also add one in the `server-vars.html`.

mrin9:
> BTW, I think we also need an example with a variable which only has the `default` and no `enum`, that should create a text `input` instead.

sure go ahead, at the moment, we don't have any automated-tests, the examples serves us our test cases, so if some thing is missing then please update the example, or if you add any new functionality, you may add new example/tests to showcase/verify the functionality

siavashs:
@mrin9 done, I updated one of the existing examples to show the cases with/without `enum`s which are rendered as `select` or `input`.
I also added the logic to set the `default` `option` to `selected`.

PR #252, opened 2020-06-28
Dual purpose download/show button

meztez:
Added recognized types to download button behavior
Slightly modify behavior for image/audio/video to display the same button with the label "SHOW" instead and open the content in a new window.

Only made the change in this file. compiled locally to test it out. Did not commit compiled files.

Putting images, videos, audio on the ui broke layout for larger image, I think it is simpler to just let the browser handle it in a new window, also cleaner looking.

Any comments?

### DIFF
--- a/docs/examples/binary-response.html
+++ b/docs/examples/binary-response.html
@@ -1,0 +1,34 @@
+<!doctype html>
+  <head>
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-132775238-1"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+
+      gtag('config', 'UA-132775238-1');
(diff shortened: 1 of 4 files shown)


mrin9:
thanks for the PR looks good,
Our current users who get binary contents (mostly zip/pdf and images) and they prefer it downloading and saving it in a location of their choice. 

For Images and PDFs it may be a good idea to allow user to open it directly using the browser in a new window, but do not want to take away the functionality of downloading it. So may be of images and pdf we can provide an additional button to show.

Also it will be great if you can add an example and a spec to demonstrate your contribution at https://github.com/mrin9/RapiDoc/blob/master/docs/list.html

you can create a simple spec for any free sample image service such as https://picsum.photos/

meztez:
yep yep, two buttons, got it.

meztez:
And done.
Are you OK with the buttons stacked on top of each other? Might be a better fit for every screen size I guess.

mrin9:
Thanks for the PR !!!
Agree, stacking button is better to accommodate more screen sizes. If we get any user feedback will see then

PR #258, opened 2020-07-06
rapidoc/components/schema: allow JSON schema content data

### LINKED ISSUE #253: Support for rendering individual JSON schema
It would be nice to support rendering vanilla JSON schema in addition to OpenAPI specs.

Looking at the implementation of `<schema-tree>` and `<schema-table>`, it seems that they currently get an intermediate representation of the schema thru their `.data` attributes.

Do you think it would be desirable to extend those components to allow passing a JSON schema thru the element body, or would you rather see another component by introduced, ex:
```
<rapi-schema>
   <pre>
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "http://example.com/product.schema.json",
  "title": "Product",
  "description": "A product from Acme's catalog",
  "type": "object",
  "properties": {
    "productId": {
      "description": "The unique identifier for a product",
      "type": "integer"
    }
  },
  "required": [ "productId" ]
}
   </pre>
</rapi-schema>
```
(using a `<pre>` would allow to fallback to just displaying the raw schema.

proppy:
This change allow usage of `<schema-tree>` and `<schema-table>` custom element with a child `<pre>` block containing a JSON schema.

Example:
```
<schema-table>
   <pre>
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "http://example.com/product.schema.json",
  "title": "Product",
  "description": "A product from Acme's catalog",
  "type": "object",
  "properties": {
    "productId": {
      "description": "The unique identifier for a product",
      "type": "integer"
    }
  },
  "required": [ "productId" ]
}
   </pre>
</schema-table>
```

Fixes #253


proppy:
- [x] example under the /docs/examples folder
- [x] link the example in /docs/list.html

> [ ] associated spec under /docs/spec folder

Is this applicable since the json schema spec is meant to be inlined?

mrin9:
I tried to test the PR, but the example you added is not working

proppy:
Seems to be related to https://github.com/w3c/webcomponents/issues/809 (children are not guaranteed to be loaded when `connectedCallback` is executed).

proppy:
PTAL 

- [x] used `<slot>` to work around the issue.
- [x] changed the implementation to expose a new `<rapi-schema>` element that wrap `<schema-tree>` and `<schema-table>` and expose properties similar to `<rapi-doc>`.
- [x] rebased against master
- [x] regenerated `dist/` and `docs`

mrin9:
will be revisiting json-schema support once OpenAPI v3.1.0 is releases and becomes little stable

onavratil-monetplus:
Any chance to revisit this topic?

PR #265, opened 2020-07-21
Make the components link clickable in the nav bar

dahu33:
This PR adds the missing `data-content-id` attributes to make the nav bar components links clickable when used with the following config:
```
<rapi-doc 
render-style="read"
show-components="true"
>
```

Thanks!


mrin9:

PR #275, opened 2020-08-04
Show navigation item as path

poadoc:
This change allows user to show navigation item as path.

Example:

```
<rapi-doc 
      show-info = "true"
      spec-url = "../specs/petstore_extended.yaml"
      show-header="false" 
      render-style="focused"
      allow-try="false" 
      regular-font="Open Sans"
      mono-font = "Roboto Mono"
      show-nav-item-as="path"
    > </rapi-doc>
```

### DIFF
--- a/docs/examples/show-nav-item-as-path.html
+++ b/docs/examples/show-nav-item-as-path.html
@@ -1,0 +1,30 @@
+<!doctype html>
+  <head>
+    <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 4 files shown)


mrin9:
Thanks for the PR, will review this soon

mrin9:
the above mentioned changes have been incorporated and the API is updated

- Path is now displayed along with HTTP Method
- attribute `use-path-in-nav-bar = (true/false)` can be used to display path or summary/description in the nav-bar

Issue #273, opened 2020-08-02
All security schemes are sent during Try regardless of spec path.security


ShaneBoyer:
First, thanks for this great project!  All security schemes are sent during Try regardless of path.security, which can result in incorrect Try requests. 

A minimal schema demonstrating this issue is below.  The `repositories-2` path uses the default security schemes, which does not include the basic scheme.  Load the schema below with allow-try on, set all three security schemes with values, then execute the repositories-2 path with the network tab open.  The Authorization header contains "Basic xyz" when it should contain the value entered for api_key1.  The `repositories` path exhibits a similar problem in that it is explicitly set to only use the `basic` scheme but trying it also sends the `specialKey` header from api_key2 even though that scheme does not apply. 

The issue appears to be that `src/components/api-request.js::onTryClick()` does not filter `this.api_keys` based on the `path.security` list, but uses all defined security schemes. 

```
openapi: 3.0.1
info:
  title: Different default and per-operation security schemes
  description: Spec with multiple security schemes, not all of which are default
  version: "1.0"
servers:
- url: https://api.bitbucket.org/2.0
security:
- api_key1: []
- api_key2: []
paths:
  /users/{username}/repositories:
    get:
      tags:
      - Repo
      description: |-
        All repositories.
      parameters:
      - name: username
        in: path
        description: |
          This can either be the username or the UUID of the account.
        required: true
        schema:
          type: string
      responses:
        default:
          description: Unexpected error.
      security:
      - basic: []

  /users/{username}/repositories-2:
    get:
      tags:
      - Repo
      description: |-
        All repositories owned by a user/team. 
      parameters:
      - name: username
        in: path
        description: |
          This can either be the username or the UUID of the account.
        required: true
        schema:
          type: string
      responses:
        default:
          description: Unexpected error.

components:
  securitySchemes:
    api_key1:
      type: apiKey
      name: Authorization
      in: header
    api_key2:
      type: apiKey
      name: specialKey
      in: header
    basic:
      type: http
      scheme: basic
```

mrin9:
Thanks for reporting.
I have added this to our issue list. I can take some help if anyone wants to contribute

mrin9:
@ShaneBoyer  thanks for the PR #277

PR #277, opened 2020-08-05
Filter security schemes based on path.security, add example spec

ShaneBoyer:
See issue #273


mrin9:
thanks for the PR, It is an important fix, I will review this soon.  !!!

Issue #291, opened 2020-08-31
feature request: bypass browser caching


ahmadnassri:
I'm using this as part of API development lifecycle (in a docker container) and it seems the browser calls for spec files are all cached by the browser:

![image](https://user-images.githubusercontent.com/183195/91774314-84a1e880-ebb6-11ea-8662-d674b639c486.png)

using it early in the dev lifecycle would need to manually resetting the cache between edits to see new results, ideally there would be a flag to add `Cache-Control: no-cache` to the requests.

Uzlopak:
Well how about using cache control html header in your html?

ahmadnassri:
I'm not sure you can actually do that from HTML ...

it would have to be a flag on the component itself, which in turn is handled by the application logic which loads the schema files

mrin9:
we will take a look into this a bit later and how others are dealing with it. But you may try to use some cache busting by using some random string  for the spec like below
https//example.com/myspec.yml?**random-12345**

ahmadnassri:
not sure how adding a random string in yaml link would help if i'm editing yaml files and using the tool for preview purposes ... that would entail manually editing the url after every edit ...

mrin9:
use JS to generate your `.yml` url

`url = 'https//example.com/myspec.yml?random-${Math.random()}' `

it shouldn't matter if you are editing your `yml` or not, every time you access it, it will give you a new url, therefore will invalidate your cache.

ahmadnassri:
I'm using the docker image to render the html... so unless I edit the

PR #294, opened 2020-09-02
add CacheControl to all requests

Uzlopak:
Always bypass Caching
making caching optional makes no sense


Uzlopak:
#291

mrin9 (approved):

mrin9:
Thanks for the PR !!!

ahmadnassri:
agreed with @mrin9

jdemaeyer:
Note that this change broke the "TRY" button for APIs that are served from a different origin unless these explicitly allow the `Cache-Control` header in their CORS preflight response. (Not a change request, just wanted to let you know :))

PR #301, opened 2020-09-14
Add support for scopes in oauth client_credentials authorization flow

JakobTeske:
With this commit, the selected scopes are as url form parameters, within the token network request, when client_credentials authorization flow is being used.

### DIFF
--- a/docs/specs/oauth2-with-scopes.yaml
+++ b/docs/specs/oauth2-with-scopes.yaml
@@ -1,0 +1,40 @@
+openapi: 3.0.0
+info:
+  title: Petstore API
+  description: Test API which supports scopes for Oauth2 Client Credentials work flow.
+  version: "1.0"
+servers:
+  - url: 'https://exampleurl.org'
+security:
+  - exampleOauth2Flow: []
+paths:
+  '/dogs':
+    get:
+      summary: Return an array of the petstore's dogs.
+      responses:
+        '200':
+          description: A successful request returns an array of dogs.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  dogs:
+                    type: array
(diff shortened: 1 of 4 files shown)


src/templates/security-scheme-template.js
```diff
@@ -45,7 +45,7 @@ function updateOAuthKey(apiKeyId, tokenType = 'Bearer', accessToken) {
 
 /* eslint-disable no-console */
 // Gets Access-Token in exchange of Authorization Code
-async function fetchAccessToken(tokenUrl, clientId, clientSecret, redirectUrl, grantType, authCode, sendClientSecretIn = 'header', apiKeyId, authFlowDivEl) {
+async function fetchAccessToken(tokenUrl, clientId, clientSecret, redirectUrl, grantType, authCode, sendClientSecretIn = 'header', apiKeyId, authFlowDivEl, scopes = null) {
```
JakobTeske, line 48: Added an optional parameter for scopes. If scopes are set, they are sent with the token request.

docs/specs/oauth2-with-scopes.yaml
```diff
@@ -0,0 +1,40 @@
+openapi: 3.0.0
```
JakobTeske, line 1: Unfortunately, I was not able to find a suitable public API, so I created this dummy API here.
I tested the functionality, by debugging the network requests and by testing against a real private API. Functionality worked, as expected.

JakobTeske:
If a scope is selected is it added to the token network request:

grafik


If multiple scopes are selected, they are added to the token network request:
grafik

If no scope is selected, there is no `scope` parameter in the network request:
grafik

mrin9:
Thanks for the PR !!!

PR #303, opened 2020-09-15
Search by params

poadoc:
This functionality gives the user a possibility to perform a search by method parameters name or description and show them methods where these parameters are used.


mrin9:
I dont see the button properly formed when I use the example that you added in `list.html`

Screen Shot 2020-09-15 at 9 44 36 AM

poadoc:
A new pull request will be created.

PR #304, opened 2020-09-15
Show only non empty component types

d-mo:
When `show-components` is true, RapiDoc will display all component types in the navbar, even when there are no components of that type available in the spec.

This PR ensures that only the types of the available components will appear in the navbar.


mrin9 (approved):

mrin9:
Thanks for the PR !!!

PR #317, opened 2020-10-07
Fix bad equality check in security scheme template

aravindanve:


mrin9 (approved):

mrin9:
Thanks much, for spotting and taking care of the issues... it saved a lot of time on our end as many of our automated tests were failing !!!

Issue #297, opened 2020-09-08
Feature request: Search by params


poadoc:
Dear developers,

It would be nice if a user can perform a search by parameters of methods (name + description). A search result can show a full description of the parameter and a list of methods where it does use.

mrin9:
to few details to properly analyze the request ... can you be little more descriptive probably with an example

poadoc:
This functionality gives the user a possibility to perform a search by method parameters name or description and show them methods where these parameters are used.
Related PR: https://github.com/mrin9/RapiDoc/pull/303

poadoc:
The PR has been updated.

PR #319, opened 2020-10-08
Add search by parameters

poadoc:
#297

### DIFF
--- a/docs/examples/search-by-params.html
+++ b/docs/examples/search-by-params.html
@@ -1,0 +1,40 @@
+<!doctype html>
+<html>
+    <head>
+        <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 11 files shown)


mrin9:
Thanks for the PR. It is a great contribution.

We are in the process of receiving feedback from our design team and user-base for the features. Also testing performance with large specs.

Following are some of the findings
- Design team wanted some minor style changes. Also there should be support for both dark and light theme. currently it supports only the light theme
- User base and User-experience have reported some tweaks to the way the search result is represented
- It should offer more flexibility in terms of what user like to search... not just parameter search
- do not search parameter descriptions/summary as some large specs gets into performance issues

But having said that, **DO NOT UPDATE** this PR. We are still reviewing this
We are more likely to accept this PR as is to record your contribution and then make some tweaks to it once we complete the feedback process

mrin9 (approved):

mrin9:
Thanks for the great PR !!!

PR #321, opened 2020-10-13
Fix example defined on array property being ignored

trebler:
This PR fixes issue when an example defined directly on `array` type object property was ignored.

### DIFF
--- a/docs/specs/examples.yaml
+++ b/docs/specs/examples.yaml
@@ -369,6 +369,18 @@ paths:
                     amount: 700,
                     currency: INR
                   }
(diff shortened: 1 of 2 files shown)


mrin9:
can you 
- add a minimal spec under `docs\specs`
- add a html under `docs\examples`
- link the html in `docs\list.html`

this should help us with testing

trebler:
@mrin9, I have added test schema to `docs/specs/examples.yaml`

mrin9 (approved):

mrin9:
Thanks for taking care of this issue !!!

Issue #322, opened 2020-10-13
Deep search in Request body


poadoc:
It would be nice if the search in a request body performed more deeply than the first level.

PR #323, opened 2020-10-13
Add deep search in Request body

poadoc:
#322

### DIFF
--- a/src/utils/common-utils.js
+++ b/src/utils/common-utils.js
@@ -53,6 +53,27 @@ export function pathIsInSearch(searchVal, path) {
   return stringToSearch.includes(searchVal);
 }
 
+export function joinObjectKeys(obj, resultStr = '') {
+  if (!obj.properties) {
+    return '';
+  }
+
+  // We should use a separated string variable in order to exclude next error 'Uncaught RangeError: Invalid string length'
+  let recursiveresultStr = '';
+  resultStr = `${resultStr} ${Object.keys(obj.properties || '').join(' ')}`;
+
+  Object.keys(obj.properties).forEach((propertyName) => {
+    if (obj.properties[propertyName].properties) {
+      recursiveresultStr = `${recursiveresultStr} ${joinObjectKeys(obj.properties[propertyName], resultStr)}`;
+    }
+    if (obj.properties[propertyName].type === 'array') {
+      recursiveresultStr = `${recursiveresultStr} ${joinObjectKeys(obj.properties[propertyName].items, resultStr)}`;
+    }
+  });
+
+  return `${resultStr} ${recursiveresultStr}`;
+}
+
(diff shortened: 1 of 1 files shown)


poadoc:
PR will be updated.

poadoc:
Workability and performance have been tested with the attached OpenAPI definition file (size: ~5MB, number of methods: 900+)
[openapi_v3.zip](https://github.com/mrin9/RapiDoc/files/5377980/openapi_v3.zip)

src/utils/common-utils.js
```diff
@@ -53,6 +53,27 @@ export function pathIsInSearch(searchVal, path) {
...
+  // We should use a separated string variable in order to exclude next error 'Uncaught RangeError: Invalid string length'
+  let recursiveStringToSearch = '';
+  stringToSearch = `${stringToSearch} ${Object.keys(properties || '').join(' ')}`;
+
+  Object.keys(properties).forEach((propertyName) => {
+    if (properties[propertyName].properties) {
+      recursiveStringToSearch = `${recursiveStringToSearch} ${processRequestBodyProperties(properties[propertyName].properties, stringToSearch)}`;
+    }
+    if (properties[propertyName].type === 'array') {
+      recursiveStringToSearch = `${recursiveStringToSearch} ${processRequestBodyProperties(properties[propertyName].items.properties, stringToSearch)}`;
```
mrin9, line 70: It seems like you are joining the string twice. the function should take an empty string initially and join it with `recursiveStringToSearch`


> I may be incorrect, just looking on a phone

mrin9 (changes requested):

src/utils/common-utils.js
```diff
@@ -53,6 +53,27 @@ export function pathIsInSearch(searchVal, path) {
   return stringToSearch.includes(searchVal);
 }
 
+function processRequestBodyProperties(properties, stringToSearch) {
```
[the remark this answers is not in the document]
poadoc: Completely agree. It has been fixed.

src/utils/common-utils.js
poadoc: if (obj.properties[propertyName].properties) {
      recursiveresultStr = `${recursiveresultStr} ${joinObjectKeys(obj.properties[propertyName], resultStr)}`;
    }
if (obj.properties[propertyName].type === 'array') {
      recursiveresultStr = `${recursiveresultStr} ${joinObjectKeys(obj.properties[propertyName].items, resultStr)}`;
}

We can have both executing of these "if" statements in the context of the one function call. Therefore we have to join string from the first 'if' statement to the second one.

mrin9 (approved):

mrin9:
Thank you for your PR !!!

mrin9 (commented):
LGTM

PR #324, opened 2020-10-16
Hide components under 'focused' template

d-mo:
"Focused" mode is great because it allows you to focus on the docs of a specific operation. However, when "show-components" is "true", all the component docs will appear at the bottom of the page, which looks rather messy.

This PR fixes that and displays the component docs only when a component has been selected.


mrin9 (approved):

mrin9:
Thank you for this great PR !!!

PR #349, opened 2020-11-16
Correct advance search to _advanced_ search

sneakyvv:
Like the title says, the label should be "advanced" instead of advance search.

I have only changed text instances, and have not touched the _allow-advance-search_ config option, since that would probably be a breaking change.


mrin9 (approved):

mrin9:
Thankyou for taking time to fix it in various places !!!

Issue #173, opened 2020-04-03
XML : missing namespace + incorrect name for array items 


bligny:
Given the following (Swagger v2) specs:

```yaml
responses:

   HeadpointPortfolio:        
    schema:
      type: array     
      xml:
        name: headpoints
        namespace: 'http://erpgis.haulogy.net/api/v1/crm'
        wrapped: true
      items:
        $ref: '#/definitions/HpPortfolioEntry'


definitions:

   HpPortfolioEntry:    
    type: object
    xml:
      namespace: 'http://erpgis.haulogy.net/api/v1/crm'
      name: headpoint          
```

The proposed xml response is:
![image](https://user-images.githubusercontent.com/19853549/78380234-d593bd80-75d3-11ea-996f-d863233ae4d5.png)

having two issues:
- xml namespaces is not rendered (red arrow)
- entries of array do not use the provided xml element name  (green arrow)

Here is the output of SwaggerUI:
![image](https://user-images.githubusercontent.com/19853549/78380115-a8470f80-75d3-11ea-881c-851d0c82a2be.png)

mrin9:
Thanks for the detailed use case and screenshot !!!
Our schema to XML converter is pretty rudimentary, 
we just started providing this feature from `7.5.0`, will be improving it soon

mohannad-musleh:
@mrin9 any updates regarding this issue?

Layap:
@mrin9 has this been fixed or something that is being worked upon? Thanks!

lothargrieb:
I changed from swagger to rapidoc. I'am very satisfied with seemless integration into our backend.
But is there any progress with xml rendering?
or an entry point to do something myself?

mrin9:
for test case checkout - https://rapidocweb.com/examples/xml.html

for using and testing on your end pick `rapidoc-min.js` from `dist`  folder in this repo

PR #350, opened 2020-11-16
Fix rendering of examples for xxx-Of schemas

sneakyvv:
Some examples are rendered weird.

#### Example schema
```
oneOf:
  - title: Schema 1
    type: object
    properties:
      foo:
        type: string
      bar:
        type: integer
    example:
      foo: foo
      bar: 42
  - title: Second schema
    type: object
    properties:
      baz:
        type: string
        format: url
```
#### Problem: 
image
* There's a `propX`
* Only an example of the first schema is shown
* Every property is using (first) `schema.example`
#### Expected: 
image
image
see https://sneakyvv.github.io/RapiDoc/examples/examples-multiple.html#get-/examples-oneOf
#### Solution
Use separate example for each schema under `oneOf`

#### Example schema
```
anyOf:
  - title: AnyOf example 1
    description: First example
    type: object
    properties:
      foo:
        type: string
      bar:
        type: integer
    example:
      foo: foo
      bar: 42
  - title: Second anyOf
    description: Another example
    type: array
    items:
      type: object
      properties:
        foo:
          type: string
        quux:
          type: array
          items:
            type: number
    example:
      foo: foo
      quux: [ 42, 24 ]
  - title: No example defined
    type: array
    items:
      type: object
      properties:
        foo:
          type: string
        quux:
          type: array
          items:
            type: number
```
#### Problem: 
image
* There's a `propX`
* Only an example of the first schema is shown
* Every property is using (first) `schema.example`
#### Expected: 
image
image
image
see https://sneakyvv.github.io/RapiDoc/examples/examples-multiple.html#get-/examples-anyOf
#### Solution
Use separate example for each schema under `anyOf`

#### Example schema
```
title: A composed object
type: object
properties:
  common:
    type: string
  other:
    type: number
anyOf:
  - title: Schema 1
    type: object
    properties:
      foo:
        type: string
      bar:
        type: integer
    example:
      foo: foo
      bar: 42
  - title: Second schema
    type: object
    properties:
      baz:
        type: string
        format: url
```
#### Problem: 
image
* There's a `propX` instead of having the properties of the `anyOf` subschemas on the same level
* Only an example of the first schema is shown
* Every property is using (first) `schema.example`
#### Expected: 
image
image
see https://sneakyvv.github.io/RapiDoc/examples/examples-multiple.html#get-/examples-anyOf-with-general-properties
#### Solution
Use separate example for each schema under `anyOf` combined with the common properties

#### Example schema
```
allOf:
  - title: Schema 1
    type: object
    properties:
      foo:
        type: string
      bar:
        type: integer
  - title: Second schema
    anyOf:
      - type: object
        properties:
          baz:
            type: string
            format: url
      - type: object
        properties:
          foobar:
            type: string
```
#### Problem: 
image
* There's a `propX`
* Only an example of the first schema is used
#### Expected: 
image
image
see https://sneakyvv.github.io/RapiDoc/examples/examples-multiple.html#get-/examples-allOf-anyOf
#### Solution
After the `allOf` has been merged, do same as with simple `anyOf`


mrin9:
awesome, thanks much for taking care of this nasty issue.

Also thanks for taking time ti documenting the PR throughly and including the screenshots and test cases !!!

mrin9:
@sneakyvv we are looking for some help to generate XML samples/examples, since you have contributed to improve examples would you like to take a stab at this issue #173

sneakyvv:
> would you like to take a stab at this issue #173

If I find the time, I'll take a look ;-)

PR #356, opened 2020-11-23
Refactor schemaToSampleObj so it returns an array of examples

sneakyvv:
As @mrin9 mentioned [here](https://github.com/mrin9/RapiDoc/pull/351#issuecomment-731655295), PR #350 broke some examples. This PR fixes those, by refactoring `schemaToSampleObj` so it returns an array of examples, and handling cases where a nested call returns multiple examples:
* for a oneOf: merge all examples for every oneOf-item
* for a anyOf: combine all examples for the regular properties with every example for every anyOf-item
* for an object: if a property has multiple examples, create an example for the object with every example of the property
* for an array: create an example with every option for the items schema


mrin9 (approved):

mrin9:
Thanks for the PR !!!

sneakyvv:
I have rebased the branch, since it had conflict on the minified js etc

sneakyvv:
I have rebased the branch because it had conflicts on the minified js etc.

PR #358, opened 2020-11-25
Add condition to protect addSchemaInfoToExample

sneakyvv:
For a schema with `"type": null`, the generated sample is "?". In that case `addSchemaInfoToExample` will cause an error in the console, since it's trying to add `::DESCRIPTION` to a string (if the schema happens to have the `description` property).

This PR fixes that by adding an early exit condition.


mrin9 (approved):

mrin9:
Thanks for fixing it !!!

PR #359, opened 2020-11-25
Use default for request input fields

sneakyvv:
When a property has no `example`, but has a `default` that default value could be taken as value for the input fields. They can be excluded of course, but imo it seems better to have a value shown in the input field whenever it's possible.


mrin9:
we have received user feedback on this and have decided not to fill `defaults`. 
Defaults should only be used for documentation rather that using it to provide any feature. It is applied by API servers when they do not receive an input.

Also certain section of users 
- like to keep the whole form empty to help them quickly enter their own data without having to remove any pre-populated values (`fill-request-fields-with-example` attribute helps you with that)
- do not want to fill values with defaults as it helps them quickly test if the API server is applying the defaults when empty parameters are provided

PR #360, opened 2020-11-25
Add null case to getSampleValueByType

sneakyvv:
The example for a schema with `"type": "null"` is rendered as `"?"`. That should be `null`.


mrin9 (approved):

mrin9:
Thanks for your PR !!!

PR #361, opened 2020-11-25
Fix expanding of paths with invalid chars

sneakyvv:
### Problem 
If you have a link for an endpoint that has a parameter, for example `foo/{id}` it would be `example.com/#get-/foo/-id-`, then in `render-style: view` the links won't expand the endpoint containers. 

### Solution
The invalid chars in the compared path in `expandTreeToPath` should also be replaced with `-`


mrin9 (approved):

mrin9 (approved):

mrin9:
Thanks... it fixed some of our test cases !!!

PR #351, opened 2020-11-17
Improve clarity of labels for anyOf & oneOf models

sneakyvv:
The labels are now using numeric keys, but if the schema has a `title` that can be used, which is clearer.

### Example schema
https://sneakyvv.github.io/RapiDoc/specs/oneof.json#/definitions/manager
& https://sneakyvv.github.io/RapiDoc/specs/oneof.json#/definitions/worker

### Current behavior
image

### Suggested new behavior
image
https://sneakyvv.github.io/RapiDoc/tests/simple-oneof.html#get-/one-of/employee

### Example schema
https://sneakyvv.github.io/RapiDoc/specs/anyof.json#/definitions/name
& https://sneakyvv.github.io/RapiDoc/specs/anyof.json#/definitions/job

### Current behavior
image

### Suggested new behavior
image
https://sneakyvv.github.io/RapiDoc/examples/anyof.html#get-/api/employee

PR #367, opened 2020-11-27
Fix title usage and array rendering in oneOf/anyOf

sneakyvv:
The labels for array items inside an `anyOf` or `oneOf`, for example with [this schema](https://sneakyvv.github.io/RapiDoc/specs/oneof-titles.json)

```
"employee": {
    "type": "object",
    [...]
    "properties": {
        [...]
        "employee": {
            "description": "Employee Details (Worker or Manager)",
            "oneOf": [
                [...]
                {
                    "type": "array",
                    "title": "An array title",
                    "description": "An array with a title",
                    "items": { "type": "string", "title": "an array item" }
                },
                {
                    "type": "array",
                    "description": "An array without a title",
                    "items": { "type": "string", "title": "an array item" }
                },
                { "type": "string", "description": "Primitive with a title", "title": "ID number" },
                { "type": "int", "description": "Primitive without a title" },
                { "type": "null", "description": "Null with a title", "title": "Null with a title" },
                { "type": "null", "description": "Null without a title" }
            ]
        },
        "anotherArray": { "type": "array", "items": { "type": "string", "title": "A string item" } },
        "objectWithArrayProp": {
            "type": "object",
            "title": "object with array",
            "properties": {
                "nestedArray": { "type": "array", "title": "A nested array", "items": { "type": "string", "title": "A nested string item" } }
            }
        }
    }
},
```

initial

was improved in #351 to **_(1)_**

current

But it is now replacing the property titles for simple arrays too **_(2)_**, because the numeric condition in `schema-table.js` was not yet optimal.
Secondly, the arrays under a `anyOf`/`oneOf` were never correctly rendered either. They were nested too deep, and rendered as type `object` **_(3)_**

This PR fixes these issues so it becomes (see commit for details how)

new

See https://sneakyvv.github.io/RapiDoc/examples/oneof-titles.html#get-/one-of/employee

### DIFF
--- a/src/components/schema-table.js
+++ b/src/components/schema-table.js
@@ -162,7 +162,7 @@ export default class SchemaTable extends LitElement {
               : html`${this.generateTree(
                 data[key]['::type'] === 'array' ? data[key]['::props'] : data[key],
                 data[key]['::type'],
-                data[key]['::title'] && !Number.isNaN(key) ? '' : key,
+                subKey || key,
                 data[key]['::description'],
                 (level + 1),
               )}`
--- a/src/components/schema-table.js
+++ b/src/components/schema-table.js
@@ -116,13 +116,13 @@ export default class SchemaTable extends LitElement {
       return html`<span class="td key object" style='padding-left:${leftPadding}px'>${key}</span>`;
     }
     let label = '';
-    let subLabel = '';
+    let optionNumber = '';
     if (key.startsWith('::ONE~OF') || key.startsWith('::ANY~OF')) {
       label = key.replace('::', '').replace('~', ' ');
     } else if (key.startsWith('::OPTION')) {
       const parts = key.split('~');
-      label = parts[1];
-      subLabel = parts[2];
+      optionNumber = parts[1];
+      label = parts[2];
     } else {
       label = key;
     }
@@ -132,7 +132,7 @@ export default class SchemaTable extends LitElement {
           ? html`
             <div class='tr ${level < this.schemaExpandLevel ? 'expanded' : 'collapsed'} ${data['::type']}' data-obj='${label}'>
               <div class='td key' style='padding-left:${leftPadding}px'>
-                ${label
+                ${label || optionNumber
                   ? html`
                     <span 
                       class='obj-toggle ${level < this.schemaExpandLevel ? 'expanded' : 'collapsed'}'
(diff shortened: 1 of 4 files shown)


mrin9:
This is breaking some of rendering. for instance 

it shows Manage and Worker as

```json
ONE OF
  1 Manager
     Manager
     Manager
  2 Worker
     Worker
     Worker  
```

instead it should be 

```json
ONE OF
  1 Manager
     name
     directReports
  2 Worker
     name
     managerName  
```

mrin9 (changes requested):
This is breaking some of rendering. for instance 
https://sneakyvv.github.io/RapiDoc/examples/oneof-titles.html#get-/one-of/employee

it shows Manage and Worker as

```json
ONE OF
  1 Manager
     Manager
     Manager
  2 Worker
     Worker
     Worker  
```

instead it should be 

```json
ONE OF
  1 Manager
     name
     directReports
  2 Worker
     name
     managerName  
```

also title rendering has been taken care of in the latest commit, you can verify at your end. 
Array rendering is still pending

sneakyvv:
> This is breaking some of rendering. for instance
> https://sneakyvv.github.io/RapiDoc/examples/oneof-titles.html#get-/one-of/employee
> 
> it shows Manage and Worker as

Oh, I didn't notice 😮 
I'll check why that's happening.

sneakyvv:
@mrin9 I've got to admit I only now took the time to properly understand what's going on in `generateTree` in `schema-table.js`.

I have renamed the variables, because, to be honest, the `prev`-prefixes threw me off. They imply that there's some complex recursive stuff happening that's doing things over 2 iterations, dealing with some of the data from the previous iteration. But... That's not the case at all.
The whole generation is actually fairly simple. 😕 😄 

Anyhow, tl;dr: it turns out that the ambiguous `subKey || key` code was not necessary, which was now apparent because of the clearer variable names (imho).

Tell me what do you think: has the code become clearer? Or am I missing some reason why the `prev` names were better and/or more correct?

You can check the same example from before: https://sneakyvv.github.io/RapiDoc/examples/oneof-titles.html

src/components/schema-table.js
```diff
@@ -107,63 +107,63 @@ export default class SchemaTable extends LitElement {
     `;
   }
 
-  generateTree(data, prevDataType = 'object', prevKey = '', prevDescr = '', level = 0) {
+  generateTree(data, dataType = 'object', key = '', description = '', level = 0) {
```
mrin9, line 110: For our internal reviewers.
In our initial implementation we used `prevkey` because it was aways a `key` we never supported title attribute. ONE OF options were also introduced later.
   
I think renamed variables makes more sense per the current functionality

mrin9 (approved):
LGTM

mrin9:
thanks for this PR !!!

PR #369, opened 2020-11-29
Bugfix/render style focused tag navigation

Dharmendra205111014:
Currently these issue when renderStyle is  focused, if Tag has description and click on tag in nav bar, it does not take to body section of the tag. if only works if tag has matching pattern of is **tag--xxx-yyy** and if name is **xxx yyy** then it breaks.


mrin9:

PR #372, opened 2020-12-01
Css selector for model type

sneakyvv:
### Problem
The XXX-OF at the top of a model table seems a bit odd.

image

### Suggestion

Add CSS selector so you can choose to style it (hide it).


mrin9 (approved):

mrin9:
Thanks for the PR !!!

PR #375, opened 2020-12-06
Bugfix/response schema fix

Dharmendra205111014:
Currently there is a bug. 

If the **responses>status>content>[content_type]** changes from one path to another it still retain the previous one and because of this it shows no schema available in read and focused render-style

Please review the change and share your opinion


mrin9:
can you please create an issue first and show how to recreate the issue, either by providing a minimal spec or  any existing example that is behaving as you explained

mrin9:
well upon looking deeper i was able to recreate the issue explained in #376 

However the fix you provided will create other issue when the response template will re-render due to reactivity and will loose its previous state. we would like to maintain its state on re-renders

I have created an alternate commit which should take care the issue you described. please take the latest build from `dist/rapidoc-min.js` and let me know if that works for your use case.

Thanks for bringing it up

PR #378, opened 2020-12-07
Bugfix/dynamic spec url change

Dharmendra205111014:
This PR belong to fix of issue #377 
The fix is made for right state of URL and selectedContentId. The other issue is anyways get addressed by this fix. 

Please review and share your views on it.

Issue #377, opened 2020-12-07
Dynamic spec-url change does not reset the states


Dharmendra205111014:
**Problem**
If you change spec-url using JS dynamically, few things are noticed

- URL still contain the old hash
- In focused render style, it again cause the issue explained in #376 because of above

**Expected behaviour**
When spec Url change, the rapi-doc should reset all states and the visible tab should be **overview** always initially for spec

To Reproduce, you can go to https://dharmendra205111014.github.io/RapiDocExamples/example1.html#post-/pet/-petId-/uploadImage and then click on change spec url button

@mrin9  please let me know your opinion on this, can open PR respectively. Thanks

mrin9:
thanks for reporting... PR is very much appreciated, 
My guess is the fix should be similar to #376

PR #379, opened 2020-12-08
- Fixed navbar active path/tag/link

Dharmendra205111014:
Hi, 

Few more fixes in continuation of issue #377 and last PR #378 

- Reset loading state when spec changes
- Reset navbar active class from path/tag/link

Signed-off-by: Dharmendra <dharmendra.patel26@gmail.com>

### DIFF
--- a/src/rapidoc.js
+++ b/src/rapidoc.js
@@ -566,19 +566,6 @@ export default class RapiDoc extends LitElement {
     }
   }
 
-  resetDoc() {
-    this.resolvedSpec = null;
-    this.loading = true;
-    this.loadFailed = false;
-    this.requestUpdate();
-
-    // Remove the previous active tag
-    const oldNavEl = this.shadowRoot.querySelector('.nav-bar-tag.active, .nav-bar-path.active, .nav-bar-info.active, .nav-bar-h1.active, .nav-bar-h2.active');
-    if (oldNavEl) {
-      oldNavEl.classList.remove('active');
-    }
-  }
-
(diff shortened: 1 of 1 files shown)


src/rapidoc.js
```diff
@@ -566,6 +566,19 @@ export default class RapiDoc extends LitElement {
     }
   }
 
+  resetDoc() {
```
mrin9, line 569: `resetDoc` name is misleading, there are many states in a the doc, it is just reseting the loader and some classes.

IMO, we do not need a separate method for this, just put these statements in `loadSpec()` method

src/rapidoc.js
```diff
@@ -566,6 +566,19 @@ export default class RapiDoc extends LitElement {
...
+    this.resolvedSpec = null;
+    this.loading = true;
+    this.loadFailed = false;
+    this.requestUpdate();
+
+    // Remove the previous active tag
```
mrin9, line 575: NavBar is available in `read` and `focused` mode only, so check if `renderStyle` is `focused` or `read` then only perform these statements

mrin9 (changes requested):

src/rapidoc.js
Dharmendra205111014: Updated PR

mrin9 (approved):
LGTM

mrin9 (approved):

mrin9:
Thanks for the PR !!!

PR #383, opened 2020-12-21
Add option to hide readOnly properties for write actions

sneakyvv:
Currently `readOnly` properties are nicely indicated by a little 🆁. However, some people may think this is confusing, as it's something that is documented but can not be provided. 
Take for example a resource's id. If you can't provide it for a `POST` request (assuming you don't use client-generated UUIDs), then it might be confusing why it's shown in the model table. Although it's technically correct, some developers might prefer to hide it.

### Suggestion

Add an optional `schema-hide-read-only` attribute on the `rapi-doc` element to determine for which write actions a `readOnly` property should be hidden.

### Examples

I have updated the [API documentation](https://sneakyvv.github.io/RapiDoc/api.html):
image

and I have added an [example html page](https://sneakyvv.github.io/RapiDoc/examples/hide-read-only.html) where you can test the different behaviors, for `POST`, `PUT` and/or `PATCH` endpoints.
(You can inspect the element to see what the JS-callbacks add on the `rapi-doc` element)


mrin9:
we did discussed this one in the past and the way we would like to approach `read` and `write` only properties is 

**ReadOnly Properties**
- Do not show them in `RequestBody` but show them in `Response`
- Example should match accordingly 

**WriteOnly Properties**
- Do not show them in `Response` but show them in `RequestBody`
- Example should match accordingly 

do not want to use a property for this one, which will help us keeping the API surface smaller and simpler.

mrin9 (changes requested):
IMO it is more appropriate to display the schema based on `RequestBody` or `Response` rather than   `API method`.
for instance irrespective of any API method,  it is always appropriate to  show `read-only` properties  if the schema is displayed in the response section.

we did discussed this one in the past and the way we would like to approach `read` and `write` only properties is 

**ReadOnly Properties**
- Do not show them in `RequestBody` but show them in `Response`
- Auto generated example should match accordingly 

**WriteOnly Properties**
- Do not show them in `Response` but show them in `RequestBody`
- Auto generated example should match accordingly 


> currently it displays read-only properties in the `schema` , but generates an appropriate example based on request-body or response

sneakyvv:
Hi @mrin9,

I understand your approach, and some openapi GUI's do it indeed like that. My change also _only_ applies to the generated HTML model (tree & table) for the `RequestBody`. It does not change anything for `Responses` nor `writeOnly` properties, although that could also be done of course.

However, I would disagree that you never want to show `readOnly` properties for `requestBodies`. Take for example a `PUT` request. In theory, this (most commonly) takes a _complete_ resource and replaces the server's representation (either immediately or delayed), even though the server can choose to ignore the `readOnly` properties, or to reject them, or to verify that they have the same value as before. Therefore, it would make sense to show those properties in the model table/tree, _if_ your server handles them properly. In other words, I believe it depends on the client's setup, and thus it would be useful if RapiDoc allowed that configuration.
As I have programmed it, it does not change the current behavior, i.e. by default the `readOnly` properties would always be shown. By taking "`always`" as default behaviour for the `schema-hide-read-only` attribute, you would achieve the behaviour you describe/prefer.

PS: I agree that `readOnly` should not be used for examples of `requestBodies`, but that already works perfectly.

mrin9:
> However, I would disagree that you never want to show readOnly properties for requestBodies. Take for example a PUT request. In theory, this (most commonly) takes a complete resource and replaces the server's representation (either immediately or delayed), even though the server can choose to ignore the readOnly properties, or to reject them, or to verify that they have the same value as before

here is how I interpret OpenAPI spec with regard to  `ignorable(optional)` and `read-only` properties 
  - `ignorable(optional)` unless you mark it `required` all properties are ignorable by server
  - `read-only` should never be sent as a part of request
  - if a property is marked as `read-only` and `required` it is applicable to `response` only

> even though the server can choose to ignore the readOnly properties, or to reject them, or to verify that they have the same value as before. Therefore, it would make sense to show those properties in the model table/tree, if your server handles them properly.

there are many instances where server wont store all the values of a request in a persistent store, they may use them for computing other value or for some verification. IMO thats something internal to the business and that shouldn't be guiding if a property should be defined as `read-only` or not.  As an API user, I do not need to know how the values are used/computed or stored internally.


> Here is what the spec say about **Read Only** properties http://spec.openapis.org/oas/v3.0.3#fixed-fields-19
Relevant only for Schema "properties" definitions. Declares the property as “read only”. This means that it MAY be sent as part of a response but SHOULD NOT be sent as part of the request. If the property is marked as readOnly being true and is in the required list, the required will take effect on the response only. A property MUST NOT be marked as both readOnly and writeOnly being true. Default value is false.

The spec clearly emphasizes that read-only property **SHOULD NOT** be sent in request.
However, I will bring up this discussion again with my extended team and our sponsoring org, and if needed will add it to our internal user survey, which is scheduled at mid of jan next year

sneakyvv:
Hi @mrin9,

Sorry, for the delayed response, but I hope you've had good holidays too 😄 

You're correct that the OpenApi specs says readOnly properties SHOULD NOT be sent, but that means that they _may_ be sent. The JsonSchema spec, which the OA spec will become [a superset of in 3.1](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#schema-object) (readOnly will no longer be described by the OA spec), says this (https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.4):

> If "readOnly" has a value of boolean true, it indicates that the value of the instance is managed exclusively by the owning authority, and attempts by an application to modify the value of this property **are expected to be ignored or rejected by that owning authority**.

So it's up to the server to throw an error or to ignore it. For this distinction to be clear, I would prefer to not show a readOnly property for a POST, which will result in an error, but still show it for a PUT, since for a PUT you _can_ send for example an id or an email address (as long as it's the same as the server has, otherwise it will also result in an error).

sneakyvv:
Hi @mrin9,

> I will bring up this discussion again with my extended team and our sponsoring org, and if needed will add it to our internal user survey, which is scheduled at mid of jan next year

Do you have any further feedback on this? (I'm using my own fork now, and would really like to move back to your official repo 😄)

mrin9:
we have scheduled a user survey within our sponsoring orgs end of this month where we included this question. once we receive some data on it will do a discussion within our team

elipinska:
Apologies for chiming in, but just wanted to say I'd LOVE to see the option to hide readOnly properties for write actions introduced to Rapidoc! I've actually recently commented on [this old issue](https://github.com/mrin9/RapiDoc/issues/43#issuecomment-766343880) to this effect, didn't realize there was already a PR for it 😍

mrin9:
We received user feedback on this and did discussed it within our team. The way we would like to address this is 
- Read-Only props should be hidden in Request
- Write-Only props should be hidden in Response

This is how most of our users model their data if they have to use it in multiple places.
The decision is also in line with how other renderers deal with this.

I have raised your points where you mentioned that `read-only` props *SHOULD NOT be sent can be interpreted as MAY BE sent* and about your reference to JSON Schema definition which states 

> attempts by an application to modify the value of this property are expected to be ignored or rejected by that owning authority

On this the team feels that in a RESTfull API context that essentially means that server must reject/ignore the `read-only` value. So if you supply a readOnly `ID` to a `PUT` method, even if the server using it for verifying that `ID` would be considered as not ignoring/rejecting it. Therefore either the spec or the implementation is incorrect.

In essence, the team thinks it is never a good idea to send readonly value in request.

Therefore we would like to modify this PR if it hides readonly values only in request (irrespective of if GET/PUT/PUT/DELETE)

sneakyvv:
Hi @mrin9,

I've finally found the time to adjust this PR. 

I also added new functionality to handle `writeOnly` properties as well. It now has the behaviour you requested, like this:

> * Read-Only props should be hidden in Request
> * Write-Only props should be hidden in Response

With one extra feature though. There still is a setting to allow developers to choose whether they want to show `readOnly` properties for certain methods (like `PUT`). Best of both worlds, right? 😃 

For `writeOnly` properties there's also a setting, but that's just an always/never setting (see https://sneakyvv.github.io/RapiDoc/api.html).

image

Check the example: https://sneakyvv.github.io/RapiDoc/examples/hide-read-only-write-only.html
All methods use the same schema for response and request, with one `readOnly` and one `writeOnly` property. See https://sneakyvv.github.io/RapiDoc/specs/readOnly-writeOnly.json:

```
{
    "type": "object",
    "properties": {
        "foo": {
            "type": "string"
        },
        "bar": {
            "type": "string",
            "readOnly": true
        },
        "baz": {
            "type": "string",
            "writeOnly": true
        }
    }
}
```

Without any settings (not touching the checkboxes), I believe it does what you described.

mrin9:
Thanks, Personally I am fine with this and I think I should be able to convince others on this as well.

Some of these functionality goes into a closed source fork of RapiDoc as it contains some domain specific functionality used by our sponsoring org, 
We tend to keep both these code base same or at least Open Sourced fork to be always a subset of closed sourced. So whatever goes into Open Source will eventually be part of closed source.
Once, I get an approval on the functionality, I will start the review process

mrin9 (approved):
Everyone is onboard with this

PR #388, opened 2020-12-29
Rename 'advance search' → 'advanced search' in code

waldyrious:
Follow-up of #349, this time with the code changes.

Note that I did rename the `allow-advance-search` config option as well, which is a breaking change. Happy to revert that change if such a compatibility break is not desired at this moment.


mrin9:
thanks for this PR, we would like to keep the code, api and ui in sync however during discussion with our release management, they would like to merge it when its closer to our release date as we have many users who uses non-released version

waldyrious:
Sure thing. Just let me know when and how I should proceed with the PR.

PR #394, opened 2021-01-21
Replace unicode return-left with something more common

jamesvl:
## Problem

The unicode character at #X2B90 does not show up in many Linux fonts, and so renders in the UI as a white rectangle (Chome) or a rectangle with odd glyphs in it (Firefox).

## Proposed Solution

This replaces the use of those codes with the similar, but far more common, #21A9.

If your browser can render them, it replaces RETURN LEFT' (U+2B90) ⮐ with LEFTWARDS ARROW WITH HOOK (U+21A9), ↩   (which Github unhelpfully converts to the emoji version, but isn't actually what you'd see in the final build)

### References

Courtesy of FileFormat.info:

* Browser [test page](https://www.fileformat.info/info/unicode/char/2b90/browsertest.htm) for 2B90
  * list of [supported fonts](https://www.fileformat.info/info/unicode/char/2b90/fontsupport.htm) (6)
* Unicode [info on 21A9](https://www.fileformat.info/info/unicode/char/21a9/index.htm)
  * list of [supported fonts](https://www.fileformat.info/info/unicode/char/21a9/fontsupport.htm) (way more than 6)

### DIFF
--- a/src/templates/header-template.js
+++ b/src/templates/header-template.js
@@ -25,7 +25,7 @@ export default function headerTemplate() {
             @change="${this.onSepcUrlChange}" 
             spellcheck="false" 
           >
-          <div style="margin: 6px 5px 0 -24px; font-size:var(--title-font-size); cursor:pointer;">&#x2b90;</div> 
+          <div style="margin: 6px 5px 0 -24px; font-size:var(--title-font-size); cursor:pointer;">&#x21a9;</div> 
         `
       } 
       ${(this.allowSpecFileLoad === 'false')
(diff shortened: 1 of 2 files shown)


mrin9:
thanks for the detailed inspection, explanation and providing a PR for it.
Am accepting it as its a better alternative, however we need to get away from os/browser dependency and i am looking at converting them to `svg` something like we did it [here](https://github.com/mrin9/RapiDoc/blob/5a6d18952fb84fb7463ce42c914bd934bc547912/src/templates/security-scheme-template.js#L384)

mrin9 (approved):
LGTM

mrin9 (approved):
LGTM

Issue #407, opened 2021-02-09
Fix empty tags in the search output


poadoc:
Hi!
In case of using search field, the navbar can contain empty tags which can annoy if we have many tags. In order to avoid a long list of empty tags in the navbar, empty tags can be skipped during the render if they don't contain any path.

mrin9:
thank you for the PR @poadoc , the feature addition have greatly improved our filter UX

PR #408, opened 2021-02-09
Fix empty tags in the search output

poadoc:
#407


mrin9:
thank you for the PR,  we received many positive feedback on this feature !!!

PR #412, opened 2021-02-10
Fix decoding

GeorgopoulosIoannis:
Windows-1253 encoding decodes to question marks

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -1211,7 +1211,7 @@ export default class ApiRequest extends LitElement {
       const contentType = tryResp.headers.get('content-type');
       if (contentType) {
         if (contentType.includes('json')) {
-          if (RegExp('charset=([^"\']+)').test(contentType)) {
+          if (/charset=[^"']+/).test(contentType)) {
             const enc = contentType.split('charset=')[1];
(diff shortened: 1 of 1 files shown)


src/components/api-request.js
```diff
@@ -1211,8 +1211,15 @@ export default class ApiRequest extends LitElement {
       const contentType = tryResp.headers.get('content-type');
       if (contentType) {
         if (contentType.includes('json')) {
-          respJson = await tryResp.json();
-          me.responseText = JSON.stringify(respJson, null, 2);
+          if (RegExp('charset=([^"\']+)').test(contentType)) {
```
waldyrious, line 1214: A couple stylistic suggestions:
- Use the inline regex constructor `/.../`, which makes this more concise and avoids the need to escaping the `'`
- Remove the capturing group `(...)`, which isn't used for anything anyway

Both of these changes IMO make the code more readable (shorter, and less of a symbol soup):

```suggestion
          if (/charset=[^"']+/).test(contentType)) {
```

mrin9:
Thanks @GeorgopoulosIoannis for the PR and thanks @waldyrious for taking time to review !!!

Issue #235, opened 2020-06-16
OAuth2 Scopes do not render


kamaydeo:
I don't know if it's a missing features or a bug but OAuth2 scopes don't show up. See `Security` section in the yaml.

```yaml
get:
      tags:
        - APIs
      summary: List all Services
      operationId: listApis
      parameters:
        - in: query
          name: page
          schema:
            type: integer
            default: 1
            minimum: 1
          description: The location of page to be searched.
          example: 5
          required: false
        - in: query
          name: size
          schema:
            type: integer
            default: 20
            minimum: 1
            maximum: 2000
          example: 100
          description: Number of records per page.
          required: false
        - in: query
          name: sort
          schema:
            type: string
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Page.GetApisResponse'
      security:
        - Bearer:
            - admin
        - OAuth2:
            - admin
```

mrin9:
Not sure of what you are expecting in the UI.
I could see the scopes in the UI under the **AUTHENTICATION** section
https://mrin9.github.io/RapiDoc/examples/oauth.html#authentication

kamaydeo:
Have a look at `AUTHORIZATIONS` section
![image](https://user-images.githubusercontent.com/2191760/84838479-8a048380-b008-11ea-95b8-9ae4829f4542.png)

mrin9:
thats not `scopes` those are the `security` needed by the operation and these `seucrity` must be defined under `security-schemes`

for more info refer to #200 on how RapiDoc shows them

kamaydeo:
I already have defined `security` and `securitySchemes`.

```yaml
security:
  - Bearer:
      - write:domain
      - admin
  - OAuth2:
      - write:domain
      - admin
```

and

```yaml
components:
  securitySchemes:
    Bearer:
      type: apiKey
      description: |
        Bearer scheme uses static perpetual token and is less secure.
        Token has a `write:domain` or `admin` permission associated with it.
        It determines whether you are authorized to invoke a particular operation or not.
      in: header
      name: Authorization
    OAuth2:
      type: oauth2
      description: |
        OAuth2 scheme is  a safer and more secure way to give you access with a short-lived token.
      flows:
        clientCredentials:
          tokenUrl: https://...../oauth2/token
          scopes:
            'write:domain': Manage Services and Subscribers in your Domain
            admin: Manage Services and Subscribers in all Domains
```

[...]

PR #422, opened 2021-02-23
#235 Display operation scopes if defined in security scheme

mb1511:
Slight UI change to display required scopes for OAuth 2.0 access tokens when defined in the operation `security` field (https://swagger.io/specification/#optional-oauth2-security) e.g.
```yaml
  security: 
    - OAuth:
      - required/scope:read
      - required/scope:write
```

Example display:

rapidoc_oauth

Related to #235 

Also added a link to the auth section, though ideally it would link to the exact auth flow.


mrin9 (approved):
LGTM.
However styling needs to get approved from UX. 
Instead of back n forth with the contributor, I suggest to merge as is to record the contribution and styling can be changed later (if needed) based on UX feedback

mrin9:
Thank you for the PR !!!

PR #434, opened 2021-03-11
Fix parsing empty responses

sneakyvv:
When a `Content-Type` header (e.g. `application/json`) is given when there is actually no content (e.g. in a `204` response), you currently get this
image
This is because it's trying to parse "nothing" as JSON, which fails. A side effect is that the response box is not updated, and still contains the response from a potential previous request.

This is fixed by checking if the response is empty before attempting (any sort of) parsing.


mrin9 (approved):

mrin9:
Thanks for digging into this and fixing it !!!

PR #437, opened 2021-03-11
Allow for relative URL's in the authentication configuration

### LINKED ISSUE #436: Relative URL's for OAuth flow dependent on selected server
Hi,

In my Open API spec, I have two servers, which are both able to handle our OAuth flow. I did set up relative url's in my OAuth configuration following the spec (https://swagger.io/docs/specification/api-host-and-base-path/#relative-urls).

The relative url's do not work though (`TypeError: Failed to construct 'URL': Invalid URL`).

It would be very nice if the selected server's URL would work when the OAuth authorization and token url's are relative URLs.

I wouldn't mind making a merge request for this if you'd be able to give me some pointers. It looks like https://github.com/mrin9/RapiDoc/blob/master/src/templates/security-scheme-template.js#L181 is the place to be to convert the relative URL to an absolute one? Any pointer on how I can find the selected server from within that method?

WouterSioen:
This will make sure that relative URL's will be resolved dependent on the
selected API server, making the authentication part work when you have
multiple API servers.

The Open API spec for relative URL's can be found here:
https://swagger.io/docs/specification/api-host-and-base-path/#relative-urls

Fixes #436

### DIFF
--- a/src/templates/security-scheme-template.js
+++ b/src/templates/security-scheme-template.js
@@ -179,6 +179,19 @@ async function onInvokeOAuthFlow(apiKeyId, flowType, authUrl, tokenUrl, e) {
 /* eslint-disable indent */
 
 function oAuthFlowTemplate(flowName, clientId, clientSecret, apiKeyId, authFlow) {
+  let authorizationUrl = authFlow.authorizationUrl;
+  let tokenUrl = authFlow.tokenUrl;
+  let refreshUrl = authFlow.refreshUrl;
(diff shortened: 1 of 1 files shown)


mrin9 (approved):
there are some minor changes, but will approve them and take care of them at our end

mrin9:
Thank you for your contribution !!!

WouterSioen:
Thanks! Any idea when we can expect a release including this change?

PR #438, opened 2021-03-11
Follow-up: Fix parsing empty responses

sneakyvv:
I noticed the fix in PR #434 wasn't correct (the wrong variable was set to empty, so the response box wasn't getting cleared, in case you already had a response from a previous request), and I already pushed extra commits to my branch, but didn't notice you already merged the other PR.

So this is a rebase of my branch.


sneakyvv:
I noticed I was just 5 minutes too late 😅 
image

mrin9:
thanks for taking care of it ...!!!

PR #443, opened 2021-03-18
Make sub titles consistent

WouterSioen:
There was only one (sub)title with a colon after it. Can we make this consistent by removing this one colon?


mrin9:
Thanks for spotting and fixing !!!

PR #455, opened 2021-03-26
Fix allowed values splitting to use ┃

wparad:
AllowedValues are separated with a `┃` not a `,`.

image


mrin9:
Thanks for the PR, this was fixed in ae406b1a303d9e0fb1c1d9e9a4dcd3c58a4f41b7 

> you didn't noticed coz it wasn't pushed as it was under testing

Issue #494, opened 2021-06-07
Feature Request - add possibility to show/hide examples


phoenixxn:
In big projects where migration to OpenAPI cannot be done in one go examples can look ugly when no real values provided, making sandbox useless (sending sandbox request wit auto-generated values is worthless).

Thus it's worth adding show-examples option, that will allow to hide/show examples for such projects.

mrin9:
please re-submit the PR as it did not expected the way it should work. could be due to accidental overwrite during resolving merge conflicts

PR #495, opened 2021-06-07
#494 Add option show-examples

phoenixxn:
#494

### DIFF
--- a/docs/api.html
+++ b/docs/api.html
@@ -441,6 +441,16 @@
             <td>false</td>
           </tr>
           <tr>
+            <td class="attr-col mono bold right">show-examples</td>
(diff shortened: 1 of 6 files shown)


mrin9:
thanks for the PR, at the moment we are on a break,  will take a look at it after july

mrin9:
@phoenixxn I had to revert back this PR as applying this was causing the examples tab to disappear in all cases

not sure if it is due to some accidental overwrite as we fixed few merge conflicts manually.
Can you please post another PR on top of the new master ? also it will be great if you can add a demo of your feature in 
https://mrin9.github.io/RapiDoc/list.html

thanks !!!

PR #513, opened 2021-07-16
Do not show read only properties in request body forms

ldlha:
Read only properties are already properly omitted in application/json view but not in application/x-www-form-urlencoded and multipart/form-data views. This branch excludes them from the generated form


mrin9 (approved):

mrin9:
Thankyou for taking care of this !!!

PR #531, opened 2021-08-18
improve components view

SoMuchForSubtlety:
* show schemas in table or tree view
 * apply search to components
 
I'm not sure if the way I filter subcomponents is ideal. Please let me know if there is a better way.


mrin9 (approved):
LGTM

mrin9:
Thanks for this one. It is really an important PR and was a long-standing ask from other users.
I am merging this as is. 

However there is one more are of improvement. That is the RequestBodies too have Schemas in them and potentially they too can be converted to schema tree.

> I'm not sure if the way I filter subcomponents is ideal. Please let me know if there is a better way.

I am pretty fine with the search

PR #534, opened 2021-08-21
Add page-direction attribute to rapi-doc html tag

taghikhani81:
I added an attribute called "page-direction" to rapi-doc tag, to be able to use both rtl and ltr directions.
Other added styles are just for preventing English texts from going right, while using rtl.

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -108,6 +108,7 @@ export default class ApiRequest extends LitElement {
           min-height:220px; 
           padding:5px;
           resize:vertical;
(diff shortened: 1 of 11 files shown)


mrin9:
Thanks for adding this awesome feature. It will help us when we get to localization.

I did notice that there are some areas to polish it further. for instance the buttons  needs margins on left instead of right.

However, I will merge it for now and you can incrementally add improved styling as and when you get time. Or we will get to it when we start work on localization !!!

PR #542, opened 2021-08-25
Add support for json encoded query parameters

jekkel:
Also added empty schema support and fixed query parameter url generation if both object typed and primitive params are present.

I know it's a bit dirty but IMO the code base is a little bit intimidating with very large files with pretty deep nesting and insanely deep nested ternary statements.

I hope I did not break anything I would appreciate constructive feedback.


mrin9 (approved):
APPROVED 

> Works fine with RapiDoc OpenSource , but breaks aws and azure integration (closed source).
IMO we can merge it to record the contribution and then add aws and azure integration story during our sprint planning

mrin9:
Thank you for the PR !!!

PR #549, opened 2021-09-13
push to browser history when navigating to target

SoMuchForSubtlety:
Push to browser history, so we can use the browser navigate forward/back functionality.

### DIFF
--- a/src/rapidoc.js
+++ b/src/rapidoc.js
@@ -830,6 +830,8 @@ export default class RapiDoc extends LitElement {
     if (!navEl.dataset.contentId) {
       return;
     }
+    // push the current URL to history before navigating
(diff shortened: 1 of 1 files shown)


mrin9 (changes requested):
to achieve this you can replace the `history.replaceState` with `history.pushState`
https://github.com/mrin9/RapiDoc/pull/549/files#diff-3ca8bb4f81354937945f23fa9c36233b7bb08dc3f7d373ffd84cca05da14812dL873

this was the case before, however most of our user feedback indicated that they prefer `replaceState` which made us change to use `replaceState` if this is needed it must be done using a flag/attribute and the default should be always `replaceState`

mrin9:
closing it due to inactivity

PR #551, opened 2021-09-13
Added generated nonce parameter. 

patrickbowie:
This is defined in the openid-connect-core 1.0 spec (https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowSteps) as optional but may be mandatory depending on grant flow.


mrin9 (approved):

mrin9:
Thank you for digging into and getting it right !!!

PR #566, opened 2021-09-23
Fix example generation for request objects without examples and mimeType

jekkel:
Example generation dereferences `mimeType` which isn't always available for request objects. Request objects now default to JSON generation since this is what the user is expected to enter anyway. Also added optional chaining operator to be safe for the future.


mrin9:
LGTM

mrin9 (approved):

mrin9:
Thank you for taking care of this. Also nice to see it is supported with a test case

PR #567, opened 2021-09-24
Add color pair for active nav items

### LINKED ISSUE #559: nav-hover-text-color is used for selected section
In modes with the sidebar, there is no dedicated `nav-selected-text-color` and `nav-hover-text-color` is used for that instead. That may result in unreadable text because of contrast, with no way to fix it.

Example color combination that is problematic:
```
theme="light"
nav-bg-color="#ffffff"
nav-text-color="#000000"
nav-hover-bg-color="#000000"
nav-hover-text-color="#ffffff"
primary-color="#000000"
```

afranke:
This is not completely working yet, but I have a hard time figuring out what’s missing.

Docs are also missing, but I wanted to get it working first.

Fixes #559

### DIFF
--- a/src/rapidoc.js
+++ b/src/rapidoc.js
@@ -112,6 +112,8 @@ export default class RapiDoc extends LitElement {
       navTextColor: { type: String, attribute: 'nav-text-color' },
       navHoverBgColor: { type: String, attribute: 'nav-hover-bg-color' },
       navHoverTextColor: { type: String, attribute: 'nav-hover-text-color' },
+      navActiveBgColor: { type: String, attribute: 'nav-active-bg-color' },
+      navActiveTextColor: { type: String, attribute: 'nav-active-text-color' },
       navAccentColor: { type: String, attribute: 'nav-accent-color' },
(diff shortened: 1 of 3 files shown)


mrin9:
thanks for attempting a fix on this, you said "This is not completely working yet"
so whats not working, do you want us to wait on this till it is fully working ?

afranke:
As that was not a breaking issue for us, we went live with it! You can see it [here](https://matrix.org/docs/api/client-server/#overview). Maybe that helps with debugging?

afranke:
The code for that page is [here](https://github.com/matrix-org/matrix.org/blob/master/rapidoc/index.html).

mrin9:
When checking your live doc, when I hover on the tags the down chevron icon gets hidden I think due to same foreground and background color can we resolve that ?

afranke:
> When checking your live doc, when I hover on the tags the down chevron icon gets hidden I think due to same foreground and background color can we resolve that ?

We probably can, but it should be in a different PR as it is unrelated to active nav items, right? It just happens to be exposed by the choice of colour combinations on our website, doesn’t it?

PR #599, opened 2021-11-04
Wait update before sending spec-loaded

onlined:
The documentation says that spec-loaded is sent after the rendering is complete, but the current implementation only requests an update, then fires the event without waiting its completion.

With this change, all the needed updates are waited until the event is dispatched.

Thank you @bhdurak for discovering the bug and helping me test the code!


mrin9 (approved):
LGTM

mrin9:
Thanks @bhdurak and @onlined for digging and fixing this one for us !!!

PR #600, opened 2021-11-04
Add dist to .gitignore

onlined:


mrin9:
for historical reasons, we keep `dist` in github as some of our clients loads it from github to test use the unreleased(beta) versions

onlined:
@mrin9 I see no `dist` directory in `master`. Are you sure things are as intended currently?

mrin9:
you are correct, lemme checkout our build pipeline whats going on. 

I am not sure how many ppl are still using `dist`, I will keep it as is for a while and see if I receive any reservations abt it, if not may be its time to remove `dist` from the repo

PR #605, opened 2021-11-14
Fix typo: charecter -> character

beatcracker:
Harmless, but annoying :nerd_face:.


mrin9:
Thanks for taking care of that annoyance !!!

PR #612, opened 2021-11-19
fixed afterPaste to combine new and old values

### LINKED ISSUE #552: pasting a value into a list field deletes any existing values
If you try to paste an additional value into a list field, you replace all existing values instead of adding an additional value.

image
<kbd>ctrl</kbd>+<kbd>v</kbd>  ->

image

BinaryWizard904:
Fixes #552 
When pasting values into a populated array field, the new values (whether one or many) will now be concatenated onto the old values.


src/components/tag-input.js
```diff
@@ -38,7 +38,16 @@ export default class TagInput extends LitElement {
   afterPaste(e) {
     const clipboardData = e.clipboardData || window.clipboardData;
     const pastedData = clipboardData.getData('Text');
-    this.value = pastedData ? pastedData.split(',').filter((v) => v.trim() !== '') : '';
+    const pastedArray = pastedData ? pastedData.split(',').filter((v) => v.trim() !== '') : '';
+    
+    if(pastedArray) {
```
mrin9, line 43: seems like `if ... else` is not as per the style guide. but otherwise fine (can be fixed at our end)

mrin9 (approved):

src/components/tag-input.js
mrin9: will fix at our end

mrin9:
Thanks for the PR !!!

PR #614, opened 2021-12-03
Persist auth to localStorage, if enabled

ruchevits:
Optionally persist authentication to localStorage.

Disabled by default, to enable:

```html
<rapi-doc 
  ...
  persist-auth="true"
></rapi-doc>
```

https://github.com/mrin9/RapiDoc/issues/615


mrin9 (approved):
LGTM

PR #635, opened 2022-01-11
fix Dockerfile

junfengye:
make it work


mrin9:
Thanks for the PR !!!

Issue #493, opened 2021-06-06
How to programmatically change the default checked status of api scopes and pkce


davidovich:
Is it possible to activate oauth scopes programmatically ? I would like to have the scopes enabled by default when I load rapidoc.

Also, in the same manner, I would like disabling the pkce checkbox. 

x-client-id and x-client-secret were added to programmatically set the security scheme, I am wondering if scopes could be managed in a similar manner, so that the radios are active - inactive.

Thanks for this wonderful project!

NOTE: I have limited HTML knowledge, but I found out that a simple getElementById('oauth2-pkce') would not work as it is part of the shadowDom. Is there  a way to interact with the shadowDOM and achieve what I want ?

mrin9:
so you need a programatic way to toggle the scopes and `PKCE` ?

davidovich:
@mrin9 yes, as the clients will need to toggle that anyway, and the backend needs three scopes, but does not support `pkce`. So that means five clicks (3 scopes, 1 pkce, and 1 auth button). The `x-client-id` and `x-client-secret` are already injected programatically thanks to you!

mrin9:
I think you should get a token urself using client-id/secret and desired scope and once you get it you can apply this token using a method.

On how to apply a token, you may checkout this example https://mrin9.github.io/RapiDoc/examples/auth.html#operations-top

PR #662, opened 2022-02-14
programmatically change the default checked status of api scopes

micheletriaca:
Hi,
first of all thanks for your awesome work!
I've added a new vendor extension `x-default-scopes` in oauth2 security scheme to permit the pre checking of required or often used scopes. I've created the pull request in case you think it may be useful

#493


mrin9 (approved):
LGTM

mrin9:
Thank you for your contribution !!!

Issue #591, opened 2021-10-21
Query parameters default value is not rendered if the value is `0` or `false`


mohannad-musleh:
When the schema includes query parameters with default value = `0` (type: integer) or `false` (type: boolean), the default value is not shown in the rendered document.

Setup details:
- Node: 14.16.0
- NPM: 7.17.0
- Bundler: vite 2.6.3
- RapiDoc: 9.1.3 (from the CDN: https://unpkg.com/rapidoc@9.1.3/dist/rapidoc-min.js)

Example:

<details>
<summary>Schema</summary>

```js
{
    "components": {},
    "info": {
        "title": "Example Schema",
        "version": "0.0.1"
    },
    "openapi": "3.0.3",
    "paths": {
        "/api/v1/example": {
            "get": {
                "operationId": "v1_example",
                "description": "Example API",
                "parameters": [
                    {
                        "in": "query",
                        "name": "boolean_query_parameter",
                        "schema": {
                            "type": "boolean",
                            "default": false
                        }
                    },
                    {
                        "in": "query",
                        "name": "integer_query_parameter",
                        "schema": {
                            "type": "integer",
                            "default": 0
                        }
                    }
                ]
            }
        }
    }
}
```
</details> 

<details>
<summary>Screenshot</summary>

![Screen Shot 2021-10-21 at 2 28 06 PM](https://user-images.githubusercontent.com/7196034/138268230-3b430218-16a9-43e1-a67e-c6dc31d79194.png)

</details>

PR #671, opened 2022-02-21
fixed Query parameters default value is not rendered if the value is 0 or false

micheletriaca:
Hi,
I've noticed that falsy default values (0, false) are not rendered both in query parameters section and in schema section

#591

### DIFF
--- a/src/utils/schema-utils.js
+++ b/src/utils/schema-utils.js
@@ -32,7 +32,7 @@ export function getTypeInfo(schema) {
     readOrWriteOnly: (schema.readOnly ? '🆁' : schema.writeOnly ? '🆆' : ''),
     deprecated: schema.deprecated ? '❌' : '',
     examples: schema.examples || schema.example,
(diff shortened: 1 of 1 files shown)


mrin9 (commented):
Can you check if we can use [?? operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) and avoid a function call ?

mrin9 (changes requested):
Can you check if we can use [?? operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) and avoid a function call ?

micheletriaca:
I've tried using the nullish coalescing operator but in this case I haven't figured out a way to exploit it. This new solution is way cleaner anyway

mrin9 (approved):
LGTM

mrin9:

PR #688, opened 2022-03-09
Add lit-element dependency

sneakyvv:
When running `npm run build` with a custom css file, this error is shown, because of the lit-element import:

```
1:1  error  'lit-element' should be listed in the project's dependencies. Run 'npm i -S lit-element' to add it  import/no-extraneous-dependencies
```

This dependency needs to be declared, and that's what this PR does.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,19 @@
 {
   "name": "rapidoc",
-  "version": "9.1.8",
+  "version": "9.2.0",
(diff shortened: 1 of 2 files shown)


sneakyvv:
A custom file should import lit and not lit-element. The documentation was still mentioning lit-element, which is fixed in https://github.com/mrin9/RapiDoc/pull/689. So this PR is closed in favor of that one.

PR #690, opened 2022-03-10
New attributes for UI control

MukeshSalaria:
Remove expand/collapse via a property setting
To add attribute called “show-section-collapse“,  default value for this attribute is false
To show expand/collapse section control has to set the value true like below.
`<rapi-doc id="thedoc" 
                  ................
                  show-section-collapse ="true" > </rapi-doc>`
 -------------------------------------------------------------------------------------------------------------------------------
 To add custom style with CSS file and a custom class on the main body
 To add custom class in the main body added attribute called "main-css-class" use it like below 
 `<rapi-doc id="thedoc" 
                  ................
                  main-css-class="custom-css-class-name" > </rapi-doc>`
                  
   --------------------------------------------------------------------------------------------------------------------------------
   To add ref of style sheet add attribute called "custom-css-file" use it like below 
    `<rapi-doc id="thedoc" 
                  ................
                  custom-css-file="custom-css-file-name" > </rapi-doc>`

### DIFF
--- a/src/rapidoc.js
+++ b/src/rapidoc.js
@@ -130,6 +130,13 @@ export default class RapiDoc extends LitElement {
       focusedElementId: { type: String }, // updating the focusedElementId will automatically render appropriate section in focused mode
       showAdvancedSearchDialog: { type: Boolean },
       advancedSearchMatches: { type: Object },
+
+      //section collapse action
+      expandCollapseSectionAction :  { type: Boolean, attribute: 'show-section-collapse' },
+      //main body class for custom css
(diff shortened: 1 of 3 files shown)


mrin9:
Thanks for the PR, I do like the idea of allowing users to inject their own classes, however I am unable to accept this PR due to two reasons.
- when we change/add/modify user facing attributes we should be updating the documentation too
- this PR makes multiple changes. the expand/collapse feature cannot be added in its current form

However, I have created a commit from your PR which provides the functionality of injecting CSS plasses
check it out here 438a35701b41d4f1c91f7ee72334389733038eed

PR #718, opened 2022-04-08
enh: allow to define request body render style

santiagodelossantos:
This PR adds a new property `requestBodyRenderStyle` in order to allow customizations on the request body section.
My proposal is to have two modes:
1. `splitted` mode, which renders two panes, one for the expected schema and other for the body input
2. `default` mode, which renders the request section on tabs

The following picture shows the `splitted` mode render:

image

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -28,6 +28,7 @@ export default class ApiRequest extends LitElement {
     this.responseUrl = '';
     this.curlSyntax = '';
     this.activeResponseTab = 'response'; // allowed values: response, headers, curl
+    this.requestBodyRenderStyle = 'default';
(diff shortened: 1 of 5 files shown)


mrin9:
thanks this may cause an issue on narrow screen and/or in view mode. in that case they should be stacked. 
can you confirm if the UI is responsive and adjusts itself per the screen factor? will be great if  you can post a screenshot

mrin9 (changes requested):
thanks this may cause an issue on narrow screen and/or in view mode. in that case they should be stacked.
can you confirm if the UI is responsive and adjusts itself per the screen factor? will be great if you can post a screenshot

Issue #620, opened 2021-12-06
Examples rendering strangely with oneOf


jcsackett:
I'm trying to use rapidoc with a spec that is using `oneOf` in a ref, and the example (a string) seems to be getting rendered as an object.

A simplified json showing the issue:
```example.json
{
  "components": {
    "schemas": {
      "Links": {
        "additionalProperties": false,
        "properties": {
          "first": {
            "$ref": "#/components/schemas/LinkProperty"
          },
          "last": {
            "$ref": "#/components/schemas/LinkProperty"
          },
          "next": {
            "$ref": "#/components/schemas/LinkProperty"
          },
          "prev": {
            "$ref": "#/components/schemas/LinkProperty"
          },
          "related": {
            "$ref": "#/components/schemas/LinkProperty"
          },
          "self": {
            "$ref": "#/components/schemas/LinkProperty"
          }
        },
        "type": "object"
      },
      "LinkProperty": {
        "oneOf": [
          {
            "description": "A string containing the link’s URL.",
            "example": "https://example.com/api/resource",
            "type": "string"
          },
          {
            "additionalProperties": false,
            "example": {
              "href": "https://example.com/api/resource"
            },
            "properties": {
              "href": {
                "description": "A string containing the link’s URL.",
                "example": "https://example.com/api/resource",
                "type": "string"
              },
              "meta": {
                "type": "string"
              }
            },
            "required": [
              "href"
            ],
            "type": "object"
          }
        ]
      }
    }
  },
  "info": {
    "title": "Example API",
    "version": "v3"
  },
  "openapi": "3.0.3",
  "paths": {
    "/foo/{bar_id}/baz": {
      "get": {
        "description": "Get a list of bazzes your foo can access based on its bar id.",
        "operationId": "getFooBaz",
        "parameters": [
          {
            "description": "The bar id of your foo.",
            "in": "path",
            "name": "bar_id",
            "required": true,
            "schema": {
              "format": "uuid",
              "type": "string"
            }
          }
        ],
        "responses": {
          "200": {
            "content": {
              "application/vnd.api+json": {
                "schema": {
                  "additionalProperties": false,
                  "properties": {
                    "data": {
                      "items": {
                        "type": "string"
                      },
                      "type": "array"
                    },
                    "links": {
                      "$ref": "#/components/schemas/Links"
                    }
                  },
                  "required": [
                    "data",
                    "links"
                  ],
                  "type": "object"
                }
              }
            },
[...]

PR #729, opened 2022-04-21
Improve rendering of oneOf schema examples

sneakyvv:
### Explanation
In `9.2.0`, if a `oneOf` schema contains an option with a `"type": "null"`, then the `response` schema wouldn't be rendered, because the sample for it couldn't be merged with the (possible) common/general properties. This was broken since the code adding those common properties was added (df594fb), in release `9.1.0`.

It has been fixed in fc49949, but not correctly. The sample for `null` would render as `"foo": { }` instead of `"foo": null`, but at least the response was rendering again (which is the issue I was experiencing too).

I already wrote a fix targeting `9.2.0`, but then I saw that, as a fix for https://github.com/mrin9/RapiDoc/issues/620, which was addressing the issue that a `string` type schema was being exploded with a property for each character of the string, the whole multi-sample rendering code was discarded to "solve" the issue.

So, I adjusted my fix to solve that string-explosion issue too.

The actual issue is that for a `string` type schema, it doesn't make sense to have common `properties` defined, so trying to add the common properties to the sample just shouldn't be done.

### Solution
I've re-added the code from `9.2.0`, which renders multiple samples and fixed that. 
I wrote the initial code able to render multiple samples 18 months ago, so I wanted to fix the addition which added the common `properties`, and have multiple samples again.

The difference is quite simple: 
1. only add the common `properties` when there are any
2. if there are common `properties`, discard any sample from a non-object schema, as such `oneOf` (sub)schema doesn't make sense anyhow
3. if there aren't any common `properties`, then we can just use the sample directly

### Examples
- I added  an example to show that senseless `oneOf` schemas are ignored (common `properties` combined with `null`, or another non-object schema): https://sneakyvv.github.io/RapiDoc/examples/oneof.html#get-/one-of/schema/senseless-options
- That one also includes an example to showcase rendering of `non-object` type schemas (at root, `null` type is ignored): https://sneakyvv.github.io/RapiDoc/examples/oneof.html#get-/one-of/schema/root
- There was already an example showcasing all kind of types of schemas as a property of an object: https://sneakyvv.github.io/RapiDoc/examples/oneof-combinations.html#get-/one-of-primitives-and-object
- Rewrote the fix for #620, the existing example works with multiple samples now: https://sneakyvv.github.io/RapiDoc/examples/oneof-with-refs.html

### DIFF
--- a/docs/specs/oneof-combinations.yaml
+++ b/docs/specs/oneof-combinations.yaml
@@ -94,21 +94,21 @@ paths:
                         description: Null with a title
                         title: Null with a title
                       - type: 'null'
-                        description: Null without a title                            
+                        description: Null without a title
                   d:
                     type: string
-  /any-of-options-with-array-as-root:
+  /one-of-options-with-array-as-root:
     post:
-      summary: ANY-OF Options with Array as root
+      summary: ONE-OF Options with Array as root
       tags:
-        - ANY-OF Root as Array / Object
+        - ONE-OF Root as Array / Object
       requestBody:
-        description: ANY OF Option where root element is an array
+        description: ONE OF Option where root element is an array
         required: true
         content:
           application/json:
             schema:
-              anyOf:
+              oneOf:
                 - title: Option 1
                   type: array
                   items:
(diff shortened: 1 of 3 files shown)


sneakyvv:
PS: I noticed that the amount of sample has been limited, which is a sensible idea, but perhaps it can be increased a bit from the 10 it is now?

mrin9:
thanks for taking care of it. I myself wanted to revisit this one, glad you did it

mrin9 (approved):

Issue #727, opened 2022-04-21
Switch read-only and write-only behaviour in callbacks and webhooks


gherkz:
Currently, RapiDoc hides read-only fields in requests and hides write-only fields in responses.

However, this behavior is the wrong way round for callbacks and webhooks, since the request is a "read" and the response is a "write". An obvious example of how this is the wrong way round would be an object with an ID field getting POSTed to a webhook:

```json
{
    "id": 123,
    ...
}
```

Under the current behaviour, the ID field is hidden in the docs for requests that gets sent to a webhook or callback. Of course we would actually send the ID field.

Redoc had the same issue: https://github.com/Redocly/redoc/issues/1720. It looks like they solved it by correcting the behavior and adding a new setting to switch back to the old behavior if desired.

If you can think of any workarounds for this in the meantime, then please do share! Looking to add a webhook to [my API](https://unused-css.com/api/v1/docs/#overview), and would prefer not to have an awkward comment in the description that says "the ID _will_ be there, even though the docs don't mention it" 😄

> PS - Related Issue https://github.com/swagger-api/swagger-ui/issues/6294

gherkz:
No problem. IMO, rapidoc is the best solution out there over redoc and swagger UI - keep up the great work.

PR #740, opened 2022-04-27
Switch readonly/writeonly visibility in webhooks and callbacks

gherkz:
Addresses #727 

I have no idea if this is the best way of doing this, and wasn't sure if it warrants a new example or not. This change _does_ appear to work correctly, but I'm not going to pretend to fully understand the RapiDoc codebase so please do check it thoroughly first!

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -69,6 +69,7 @@ export default class ApiRequest extends LitElement {
       schemaDescriptionExpanded: { type: String, attribute: 'schema-description-expanded' },
       allowSchemaDescriptionExpandToggle: { type: String, attribute: 'allow-schema-description-expand-toggle' },
       schemaHideReadOnly: { type: String, attribute: 'schema-hide-read-only' },
+      schemaHideWriteOnly: { type: String, attribute: 'schema-hide-write-only' },
       fetchCredentials: { type: String, attribute: 'fetch-credentials' },
 
       // properties for internal tracking
@@ -322,8 +323,8 @@ export default class ApiRequest extends LitElement {
           serializeStyle || 'json',
           '',
           '',
-          true,
-          true,
+          !(this.schemaHideReadOnly && (this.schemaHideReadOnly.includes(this.method) || this.schemaHideReadOnly === 'true')),
+          !(this.schemaHideWriteOnly && (this.schemaHideWriteOnly.includes(this.method) || this.schemaHideWriteOnly === 'true')),
           'text',
           false,
         )[0].exampleValue;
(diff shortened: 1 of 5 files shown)


mrin9:
Will look into it soon, 
this will break one of our following API behavior 

- `schema-hide-read-only` : here user can provide the method name such as `POST` , `PUT`,  `DELETE` or `GET` to specify where the `read-only` fields should be hidden

but we discussed internally that we would like to move away from this and be inline with the understanding that
- Hide ReadOnly for Requests 
- Hide WriteOnly for Responses
- Opposite is true in case of Web-Hooks and Callbacks 

CC @sneakyvv for any opinion on it as he introduced `schema-hide-read-only`  attribute to RapiDoc

sneakyvv:
Hi @mrin9,

Let me quickly refer to the PR where this behavior was introduced: https://github.com/rapi-doc/RapiDoc/pull/383

I still feel that Rapidoc should allow to show the `readOnly` properties in `PUT` requests, to be completely in line with [the JSON Schema spec](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-9.4) :
> If "readOnly" has a value of boolean true, it indicates that the value of the instance is managed exclusively by the owning authority, and attempts by an application to modify the value of this property are expected to be ignored or rejected by that owning authority.

As the default behaviour of RapiDoc is to never show them for requests (i.e for the write methods, `POST`, `PUT` & `DELETE`), it's already in line with your interpretation of the spec. But again, a literal interpretation does not say that `readOnly` properties cannot be sent. 
The idea is that a resource retrieved via `GET` should be allowed to be sent with `PUT` without having to remove the `readOnly` properties. Technically the JSON schema (validation) will still allow them since in the actual schema they are present, only marked as `readOnly`. It's only the documentation which isn't showing them.
Furthermore, in combination with `"additionalProperties": false`, it might even be weirder. Although that isn't really shown in the rendered schema, it would imply that you can't send the `readOnly` properties, which simply isn't true. Only "attempts to modify the value [_should_] be ignored or rejected".

So, my proposal would be to not touch the current behaviour, but improve upon it for web-hooks and callbacks. I haven't really looked at the code changes in this PR, but at first glance it seems to do just that. If I find the time, I'll try to look at it soon.

sneakyvv:
@gherkz 

> and wasn't sure if it warrants a new example or not

An extra example is always useful.
@mrin9 Is there perhaps already one using webhooks & callbacks?

mrin9:
For now I am merging this as the solution is more closer to the spec
@sneakyvv we can revisit this if it needs to change it

gherkz:
Sorry I wasn't very responsive on this issue - was away on vacation. Perhaps https://github.com/rapi-doc/RapiDoc/issues/727 can be closed now if you're happy the solution is complete

PR #783, opened 2022-07-12
Accessibility improvements

madelinekosse:
This resolves two issues encountered when using a screen reader to read the generated documentation pages:

* The `select` elements on the response panel lacked accessible names
* The tables were being read as tabular data containing empty cells


mrin9:
We do have a plan to look at Accessibility improvements, including enabling keyboard navigation, however not able to carve out time for it. This will be a start.

Thanks much for your contribution towards this aspect !!!

PR #808, opened 2022-08-30
added support int32 and int64 to schema and support paths with spaces…

Alexeyca:
… correctly

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -1150,7 +1150,7 @@ export default class ApiRequest extends LitElement {
     const queryParamObjTypeEls = [...requestPanelEl.querySelectorAll("[data-ptype='query-object']")];
     const headerParamEls = [...requestPanelEl.querySelectorAll("[data-ptype='header']")];
     const requestBodyContainerEl = requestPanelEl.querySelector('.request-body-container');
-    fetchUrl = this.path;
+    fetchUrl = this.path.replaceAll(' ', '');
     const fetchOptions = {
       method: this.method.toUpperCase(),
     };
--- a/src/utils/schema-utils.js
+++ b/src/utils/schema-utils.js
@@ -76,6 +76,11 @@ export function getTypeInfo(schema) {
       constrain += `Max ${constrain ? '┃' : ''}${schema.maxLength} chars`;
     }
   }
+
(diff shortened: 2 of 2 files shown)


src/components/api-request.js
```diff
@@ -1150,7 +1150,7 @@ export default class ApiRequest extends LitElement {
     const queryParamObjTypeEls = [...requestPanelEl.querySelectorAll("[data-ptype='query-object']")];
     const headerParamEls = [...requestPanelEl.querySelectorAll("[data-ptype='header']")];
     const requestBodyContainerEl = requestPanelEl.querySelector('.request-body-container');
-    fetchUrl = this.path;
+    fetchUrl = this.path.replaceAll(' ', '');
```
mrin9, line 1153: `space` is a valid character in url paths. they shouldn't be modified or removed

src/utils/schema-utils.js
```diff
@@ -76,6 +76,11 @@ export function getTypeInfo(schema) {
       constrain += `Max ${constrain ? '┃' : ''}${schema.maxLength} chars`;
     }
   }
+
+  if (dataType.match(/integer|number/g) && schema.format) {
+    info.type = schema.format;
```
mrin9, line 81: `type` and `format` are two distinct attributes. we use `type` in many places to make decision on rendering complex schemas. assigning `format` to `type` is technically incorrect and also breaks  many of our rendering logic

src/components/api-request.js
Alexeyca: can you provide an example where you have URL with useful spaces

src/utils/schema-utils.js
Alexeyca: can you render LONG in SCHEMA correctly and consistently?

src/components/api-request.js
mrin9: URLs are defined in [RFC 3986](http://www.ietf.org/rfc/rfc3986.txt) . Which states that spaces and few other control characters should be percentage encoded. In case of RapiDoc this is automatically done by the browser. 
Further to this, RapiDoc should not be in the business of modifying the APIs in anyways. If the spec has a mistake they must be corrected within the spec

PR #816, opened 2022-09-06
Add support for `const`

sneakyvv:
This is allowed in OA 3.1 (See https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-6.1.3), but is not rendered properly (as a regular string for example, without the allowed values, or rather fixed value.

This PR adds support for that.

Examples: at https://sneakyvv.github.io/RapiDoc/examples/OA-3.1.html

image

image

For schema at https://sneakyvv.github.io/RapiDoc/specs/oa-3.1.json

Note: it doesn't check whether the openapi version in the schema is >= 3.1.0

### DIFF
--- a/docs/examples/OA-3.1.html
+++ b/docs/examples/OA-3.1.html
@@ -1,0 +1,47 @@
+<!doctype html>
+<html>
+  <head>
+    <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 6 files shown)


mrin9:
Thanks @sneakyvv , nice one to fix. I will review it next week

mrin9 (approved):
LGTM

mrin9:

PR #849, opened 2022-11-29
fix non minified invalid rapidoc.js build

sroucheray:
Built Rapidoc file `rapidoc.js` was invalid due to a concatenation not ending a comment


mrin9:
Thanks for your contribution !!!

PR #850, opened 2022-11-29
fix(api-request): fix not filled example value when it is 0

sroucheray:
When a parameters has an example of value 0, it was not filled by default
image


mrin9 (commented):
LGTM

mrin9:

PR #851, opened 2022-11-29
feat(curl): optionally display cURL snippet before try

sroucheray:
This PR had the capability to show cURL snippet before clicking on the TRY button.

- By default the behaviour of displaying the cURL tab remains the same.
- I added an attribute `show-snippet` to the component accepting as a value `default` | `before-try`

When setting `before-try` the cURL snippet box is displayed after the request body and before the response. In this case the cURL tab in the response is not shown.

The cURL snippet is updated when user changes values in forms.

image


mrin9 (commented):
I see only `showSnippet` attribute being added, but don't find the code which is using it.
Can you re-check the PR on your end 

also if you are modifying the PR, can you rename the attribute `showSnippet` to  `showCurlBeforeTrye : true | false` .
All our attributes starting with `show-` or `allow-` are of boolean type

sroucheray:
Sorry I messed with splitting my branch by commits. I just pushed force the changes, taking into account your feedbacks. It is all fixed now.

mrin9:
I see that you have changed the attribute name to `show-curl-before-try`, but I still dont see the code that uses its value

sroucheray:
I am not sure why you don't see changes but they are visible here https://github.com/rapi-doc/RapiDoc/pull/851/files, this is mainly 
- `api-request.js` https://github.com/rapi-doc/RapiDoc/pull/851/files#diff-8e0994f875e96985adc7136b31e1eb6fe2ba0b4ec504a11d5bfa5044d33fe600
- `rapidoc.js` https://github.com/rapi-doc/RapiDoc/pull/851/files#diff-3ca8bb4f81354937945f23fa9c36233b7bb08dc3f7d373ffd84cca05da14812d
- `endpoint-template.js` https://github.com/rapi-doc/RapiDoc/pull/851/files#diff-ce0fe146b9c1b61b44b0b490e8074dc734eebd8371304d94601c6a99c096ea33
- `expanded-endpoint-template.js` https://github.com/rapi-doc/RapiDoc/pull/851/files#diff-956c80657ef76350f5032e3257711fff3c7d9911cfe31bff8f756f779c6d4acb
- I added an API documentation example

image

sroucheray:
Hi @mrin9,

Sure I can revert the text wrapping to the previous behavior. I'll wait for all your feedback before pushing a fix.

mrin9:
Thanks for contributing to include this awesome feature, also nice work to update documentation and fix some of the typos in code  !!!

guideloom:
Need this feature. When will it be available for general use?

sroucheray:
@guideloom the attribute to set to true is `show-curl-before-try` https://rapidocweb.com/api.html

prassie:
@sroucheray 
I think @guideloom is asking for an option just to show cURL without the option to invoke/try the API, like `show-curl-without-try`.
I guess `show-curl-before-try` is functioning more like `show-curl-before-invoking-try`.


From: https://rapidocweb.com/examples/api-demo.html
image

guideloom:
Yes, that's exactly it. Would be great to just have the "curl" example showing, without the try button. I had assumed by using the following combination of options that this would be achievable:

`allow-try : false`
`show-curl-before-try: true`

However, it seems the curl display is completely dependent of "allow-try" being set to true.

sroucheray:
@guideloom I suggest you open a new ticket for that for better discussions (I am not the maintainer though :) ).

PR #852, opened 2022-11-30
fix replace history state when using deep path URL

sroucheray:
Replacing history state did not work when RapiDoc component is delivered in a deep URL web page (i.e. `http://www.mysite.com/deep/link/rapidoc`).

Rapidoc calls on `history.replaceState` did not always use the full URL (was not DRY). So in some cases we ended up with URLs like `http://www.mysite.com#overview` instead of `http://www.mysite.com/deep/link/rapidoc#overview`.


mrin9:
in the deep-link example you provided above for the reason of this PR 
can you try setting `routePrefix = '/deep/link/rapidoc#'` and check if it takes care of your situation ?

sroucheray:
I made some changes to my PR. Following your explanations I tested it more thoroughly in a combination of many cases:
- render style: view, read, focused
- with or without route prefix
- within a deep path or not

I think this PR fix also small bugs where RapiDoc does not  take into account the hash, for example here [https://rapidocweb.com/examples/read-mode.html#get-/store/order/-orderId-](https://rapidocweb.com/examples/read-mode.html#get-/store/order/-orderId-) (the hash is changed when the page loads at least in Chrome on Linux).

With this PR, the ID of an element to open/scroll to is extracted from the URL whether it is in a hash or in the URL path with a routePrefix (which was not the case).

I made a lot of tests manually. I think for the sake of the project it would be good to have unit tests to avoid regression and make sure I don't introduce one :)

mrin9:
thank you  for the contribution, another nice feature and code cleanup !!!

PR #862, opened 2022-12-13
Feat/typescript migration

sroucheray:
Hi @mrin9,

So here is the branch where I migrated the project to Typescript.
- [X] Migrate all files from JS to TS
  - [X] Use [openapi-types](https://www.npmjs.com/package/openapi-types) lib wherever possible
  - [X] Add extra typings where it was feasible
  - [X] Cast to typing where some type incompatibility may arise 
  - [X] When I had not other way I added some `any`s where typing were difficult to understand
  - [X] Try NOT to fix code yet, unless typings made it impossible to compile 
  - [ ] Migrate rapidoc-mini.js (commented out for now to avoid extra work)
- [X] Update tsconfig and webpack to adapt the build pipeline (commands did not change)
  - [X] build passes and produces artefacts (js and typing files)

The remaining tasks I see
- [ ] Improve / Cleanup custom typings by someone who has more knowledge about the shape of object
  - [ ] Cleanup code wherever typings alert
- [ ] Testing 
- [ ] Rebase on master and integrate last modification
- [ ] Simplify and make some portions of code more robust which should be much easier by the help of typings
- [ ] Add unit testing

One note about functions. You'll see that many functions have a first parameter names `this`, it's not a real parameter but the way typescript make it clear on which type of object the function is called (as many utility functions are called with `call`).
Example
```typescript
...
export default function jsonSchemaViewerTemplate(this: RapiDocJSONSchemaViewerElement, isMini = false) {
...
```


sroucheray:
Side note: my last commit is a fix for the response status not properly set when an error occurred. I encountered this error while working on a specific API. I took the opportunity to narrow down the type of the `responseStatus` to `success` | `error` instead of a string. Typescript makes it clear what this property can hold as a value now !

sroucheray:
I merged master in my branch to keep up to date with latest commits

ricodealma:
Waiting for release.

sroucheray:
> Waiting for release.

Unfortunately, the main author did not give much interest in this branch since the MR. I tried to maintain it in phase with master as much as I can, though.

PR #863, opened 2022-12-13
Use baseUrl of apiServer to calculate relative refreshUrl, tokenUrl, and authorizationUrl

### LINKED ISSUE #580: Relative url for the security schemes
Try to use RapiDoc in a new project and see some strange behavior when working with relative urls. 

Consider this openapi spec 
```
servers:
  - url: https://api.example.com/v2
components:
  securitySchemes:
    openId:
      type: openIdConnect
      openIdConnectUrl: /.well-known/openid-configuration
```
According to https://swagger.io/docs/specification/authentication/openid-connect-discovery section "Relative Discovery URL". 

> Relative URLs are resolved according to RFC 3986. In the example above, it will be resolved to 
> https://api.example.com/.well-known/openid-configuration.

Swagger-ui in the authorization section uses this rule. 
RapiDoc takes full server url and resolves  openIdConnectUrl as `https://api.example.com/v2/.well-known/openid-configuration`. 

Is this intended behavior or bug ?

jak103:
Should fix #580 Relative url for the security schemes

### DIFF
--- a/src/templates/security-scheme-template.js
+++ b/src/templates/security-scheme-template.js
@@ -264,9 +264,7 @@ function oAuthFlowTemplate(flowName, clientId, clientSecret, securitySchemeId, a
   const isUrlAbsolute = (url) => (url.indexOf('://') > 0 || url.indexOf('//') === 0);
   // Calculcate base URL
   const url = new URL(this.selectedServer.computedUrl);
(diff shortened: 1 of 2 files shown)


mrin9:
Thankyou for this awesome contribution !!!

PR #868, opened 2022-12-30
Adapt to the situation where there is a query in the path

Handgrip:
Well, although this situation is against the specification, I added this function for the program to be more robust, and in another UI(https://editor.swagger.io/), this situation is adapted.

I made as little change as possible, thank you!

Test case:
```json
{
    "paths": {
        "/proxy?interface=A": {
            "get": {
                "parameters": [
                    {
                        "name": "id",
                        "schema": {
                            "type": "integer"
                        },
                        "in": "query"
                    }
                ]
            }
        }
    }
}
```

Before:
image

After:
image


mrin9:
Thank you for your contribution , it indeed improves handling certain cases !!!

PR #870, opened 2023-01-03
Fix incorrect credentials handling for client_credentials grant type

### LINKED ISSUE #869: Incorrect setting of credentials while handling client_credentials grant type
In case of `client_credentials` grant type, when the credentials are sent in the `Authorization` header, 
current implementation (9.3.4 beta) does not set `clientId` and `clientSecret` properly.
While debugging the token request from our authorization server, the received token is `null:null`.

I'm thinking that there are some side effects when resolving #758 which causes this issue.

Kieun:
Fixes #869


mrin9:
Thanks for spotting the typo and fixing it !!!

PR #876, opened 2023-01-12
Correct HTTP basic auth header value

### LINKED ISSUE #875: Basic auth broken on unreleased master
The change in https://github.com/rapi-doc/RapiDoc/commit/26f1db71bae22820b7027c0f3c3411f6adc96921 seems to have accidentally broken basic auth as the header value is now missing the `Basic ` prefix.

https://github.com/rapi-doc/RapiDoc/blob/master/src/templates/security-scheme-template.js#L19

chadlwilson:
- fixes #875 

Looks like regression accidentally introduced with https://github.com/rapi-doc/RapiDoc/commit/26f1db71bae22820b7027c0f3c3411f6adc96921


mrin9:

chadlwilson:
Thanks for taking a look 🙏

PR #892, opened 2023-02-15
New consolidated example demonstrating various RapiDoc API attributes and events

prassie:
### Notes
- New example demonstrating most of the attributes and events of RapiDoc API
- Highly useful as a single example to toggle various options and see their effects immediately
  - Included all the options demonstrated in Play Ground example
- Single file HTML can be copied and used by the API developers to serve their own API specs

### Displaying Endpoint Summary vs Path
image
image

### Dynamically updating heading text and logo's based on Spec
image
image


mrin9:
Thank you for adding this example to showcase RapiDoc features

PR #889, opened 2023-02-08
Distinguish nullable objects when schema-style is tree

HansAarneLiblik:
Fixes [!858](https://github.com/rapi-doc/RapiDoc/issues/858)

Visuals before
image

Visuals now: 
image
image

PR #919, opened 2023-04-18
Distinguish nullable arrays when schema-style is tree

HansAarneLiblik:
Follow-up to https://github.com/rapi-doc/RapiDoc/pull/889

Also display `null|..` for nullable arrays (of arrays / objects)

Can't display `null|...` for primitives, as that is handled very differently and `nullable` is not available when rendering them

Previously: 

image

Current:

image

### DIFF
--- a/docs/specs/temp.yaml
+++ b/docs/specs/temp.yaml
@@ -28,10 +28,10 @@ paths:
           application/json:
             schema:
               type: object
-              additionalProperties: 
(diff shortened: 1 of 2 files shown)


HansAarneLiblik:
This won't work without some changes in the `generateTree` signature. Arrays are rendered, when handling children, and we don't have array `nullable` values on that level anymore

PR #941, opened 2023-07-12
Small quality of life fixes

sanscontext:
This PR:
- ✔️ Adds a link to the `#auth` anchor on the "no api key set" message, so that readers get a CTA how to fix it.
  **After**: (I don't have a non-edited version to screenshot for a Before)
  image

- ✔️ Renders the `bearerFormat` string as the field hint text, for HTTP Bearer type auth, if one is set. (If not set, uses the original `api-token`.)
  **Before**:
  image
  **After**:
  image

- ✔️ Similarly, renders the `bearerFormat` string as the noun in the instructional text, if set. (If not set, uses the original "Token String".)
  **Before**:
  image
  **After**:
  image

- ⚠️ Attempts to check if there are >1 of a type of security scheme, and add the `${v.securitySchemeId}` if there are, except I couldn't get this working or troubleshoot effectively with my rudimentary JS skills.
  Not working right now, but when it's done it would look something like this:
  **Before**: two different auth formats with the same name, so you have to use the descriptive text to tell them apart.
  image
  **After**: much easier to see that these are different
  image

Thanks all!

### DIFF
--- a/src/templates/security-scheme-template.js
+++ b/src/templates/security-scheme-template.js
@@ -401,7 +401,7 @@ export default function securitySchemeTemplate() {
           <div class="blue-text"> ${providedApiKeys.length} API key applied </div>
           <div style="flex:1"></div>
           <button class="m-btn thin-border" part="btn btn-outline" @click=${() => { onClearAllApiKeys.call(this); }}>CLEAR ALL API KEYS</button>`
-        : html`<div class="red-text">No API key applied</div>`
+        : html`<div class="red-text">No API key applied - <a href="#auth">Apply one here</a></div>`
       }
     </div>
     ${this.resolvedSpec.securitySchemes && this.resolvedSpec.securitySchemes.length > 0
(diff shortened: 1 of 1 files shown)


AdrianMachado:
Screenshots of before and after would be appreciate @sanscontext  :)

AdrianMachado:
cc @mrin9

sanscontext:
Thanks @AdrianMachado , I added some before/after screenshots, trying to limit it to just what I've touched in this PR.

I was thinking about the `bearerFormat` part, and I'd actually want to see some examples of how it's used in the wild before assuming it'd be helpful here. The OAS isn't very prescriptive on how this should be used, and "Hint" isn't going to mean the same thing to everyone.  Would this be something we could poll the Discord for, for examples?

src/templates/security-scheme-template.js
```diff
@@ -401,7 +401,7 @@ export default function securitySchemeTemplate() {
           <div class="blue-text"> ${providedApiKeys.length} API key applied </div>
           <div style="flex:1"></div>
           <button class="m-btn thin-border" part="btn btn-outline" @click=${() => { onClearAllApiKeys.call(this); }}>CLEAR ALL API KEYS</button>`
-        : html`<div class="red-text">No API key applied</div>`
+        : html`<div class="red-text">No API key applied - <a href="#auth">Apply one here</a></div>`
```
mrin9, line 404: `#auth` wont be available always.it depends on which section the user is showing and also special care must be taken for focused view

mrin9 (changes requested):

sanscontext:
Thanks all, I'm open to additions to this PR. As I mentioned to @AdrianMachado, I'm not a developer so my changes here are rudimentary but high-impact on our end. It'd be nice not to have to add them to every new version of Rapidoc I pull down.

AdrianMachado:
I can take a look later this week

sanscontext:
Hey folks, it's been six months. Should I just close this?

sanscontext:
Hey folks, is there any way to get this looked at please?

PR #949, opened 2023-08-13
Add preview of images

j3rem1e:
If the response is of type image/*, then a preview of the image is shown (like swagger-ui).

screen


mrin9 (approved):

mrin9:
Thanks for adding this feature !!!

PR #953, opened 2023-08-19
Support for x-displayName for tags

j3rem1e:
Allow to change the display name of tags without changing the "name" property. It's the same property/behavior as in redoc.


mrin9 (approved):

mrin9:
@j3rem1e Thanks for the PR. Apologies for the delay in accepting it. We are resuming our work on this library

PR #975, opened 2023-11-14
Update curl examples to include --compressed flag

rdlrt:
A very simplistic optimisation to include `--compressed` argument to generated curl examples. While this may be blocked on some  configurations, it should work for most. Especially when working with large results, this can heavily save data size being transferred over network.

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -1619,7 +1619,7 @@ export default class ApiRequest extends LitElement {
       curlUrl = fetchUrl;
     }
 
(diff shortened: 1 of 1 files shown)


mrin9:
The curl syntax is provided solely to represent the equivalent API call. The actual API request is made using JavaScript's Fetch API, not curl. Since the purpose of the curl syntax is to illustrate the API call, including --compressed  flag in syntax may cause confusion, as it might imply that special handling is required to manage compressed responses

PR #984, opened 2023-12-10
Fix #966: Search click handler on spans

### LINKED ISSUE #966: Advanced Search endpoint click events not registering on text, but do register in empty space.
It seems there might be some sort of click propagation or z-index issue with advanced search. Clicking on any of the endpoint's text swallows the event, but clicking in empty space (either to the right or just above/below text) will successfully route you to the endpoint.

https://github.com/rapi-doc/RapiDoc/assets/6569636/b1272e39-746d-4e7c-b30a-90fb563306ab

aalshakhshir1:
Fixes an issue (https://github.com/rapi-doc/RapiDoc/issues/966) with search where clicking on endpoint path/method is not scrolling due to event handler checking on event.target.contentId attribute. open to ideas/thoughts for a better fix

    if (!navEl.dataset.contentId) {
      return;
    }

This fix makes sure when clicking on any child element to scroll to that endpoint

element: https://github.com/rapi-doc/RapiDoc/blob/18afdce428b1bc6838d52c9f391a982817271da4/src/templates/advance-search-template.js#L48

**Event handler condition** 

https://github.com/rapi-doc/RapiDoc/blob/18afdce428b1bc6838d52c9f391a982817271da4/src/rapidoc.js#L952

**Video:**

https://github.com/rapi-doc/RapiDoc/assets/139879471/9e49658c-6632-4aa5-ace7-4cae272fcee6

### DIFF
--- a/src/templates/advance-search-template.js
+++ b/src/templates/advance-search-template.js
@@ -59,9 +59,9 @@ export default function searchByPropertiesModalTemplate() {
           }
         }"
       > 
-        <span data-content-id='${path.elementId}' class="upper bold-text method-fg ${path.method}">${path.method}</span> 
-        <span data-content-id='${path.elementId}'>${path.path}</span>
-        <span data-content-id='${path.elementId}' class="regular-font gray-text">${path.summary}</span>
+        <span style="pointer-events: none" class="upper bold-text method-fg ${path.method}">${path.method}</span> 
+        <span style="pointer-events: none">${path.path}</span>
+        <span style="pointer-events: none" class="regular-font gray-text">${path.summary}</span>
       </div>
     `)
     }
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -1465,6 +1465,21 @@ export default class ApiRequest extends LitElement {
       const startTime = performance.now();
       fetchResponse = await fetch(fetchRequest, { signal });
       const endTime = performance.now();
+      // Allow to modify response
+      let resolveModifiedResponse; // Create a promise that will be resolved from the event listener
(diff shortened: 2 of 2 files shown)


Aalshakhshir:
Hello @mrin9 

Any updates on this one please

src/templates/advance-search-template.js
```diff
@@ -59,9 +59,9 @@ export default function searchByPropertiesModalTemplate() {
           }
         }"
       > 
-        <span class="upper bold-text method-fg ${path.method}">${path.method}</span> 
-        <span>${path.path}</span>
-        <span class="regular-font gray-text">${path.summary}</span>
+        <span data-content-id='${path.elementId}' class="upper bold-text method-fg ${path.method}">${path.method}</span>
```
mrin9, line 62: instead of adding `data-content-id='${path.elementId}'` can you try by adding `style="pointer-events: none"`
`Id`s in general shouldn't be duplicated.

mrin9 (changes requested):
Thanks for the PR. 
And I am very sorry to reply late on this. 

We are starting to resume our on this library. Can you see if you can make the change I requested in the comments

src/templates/advance-search-template.js
aalshakhshir1: Thank you @mrin9 , makes sense , that worked as well , will push it soon.

Thanks again.

aalshakhshir1:
Done, thanks @mrin9 🙏

mrin9 (approved):

mrin9:

PR #1019, opened 2024-07-25
fix: adding form field element ids

### LINKED ISSUE #1018: add form field element with id or name attribute
There are some form field element has neither an id nor a name attribute.
This might prevent the browser from correctly autofilling the form.

There should be a unique id or name attribute added to the various form fields.
This is not strictly needed, but would be nice to have on for instance advanced-search dialog options.

tobiasehlert:
This PR is for adding missing IDs to some of the inputs (and textareas).

fix #1018


mrin9 (approved):

mrin9:
thanks a lot for the contribution !!!

Issue #1023, opened 2024-09-03
CURL command displaying incorrectly.


BobKurtis:
I'm new to rapi-doc but I've run into an issue that I can't seem to find an answer for. It seems if I supply the "accept" header and then supply the content response type. The curl command that is generated has the accept value there twice

```
/end-point-example:
    get:
      tags:
        - Example
    .....
        - name: Accept
          in: header
          description: |
            Specifies the response format that the client would like to receive back. This must be `application/json`.
          example: 'application/json'
          required: true
          schema:
            type: string
.....
      responses:
        '200':
          description: OK
.....
          content:
            application/json:
```

and this leaves my curl command looking like

```
curl -X GET "exampleURl.com" \
 -H "accept: application/json, application/json"\
 -H "accept-encoding: gzip"\
 -H "user-agent: TravelNow/3.30.112" \
```
Not sure what I'm doing wrong, any guidance would be greatly appreciated.

mrin9:
@BobKurtis thanks for the contribution

PR #1024, opened 2024-09-05
Curl header dedupe

BobKurtis:
This removes duplicate headers found in generated cURL examples. This address the issue https://github.com/rapi-doc/RapiDoc/issues/1023 . This really only effects the "accept" headers but is written in such a way that all duplicates are removed.

Before:
`curl -X GET "https://example.com" \
 -H "accept: application/json, application/json"\
 -H "accept-encoding: gzip"\
`

After:
`curl -X GET "https://example.com" \
 -H "accept: application/json"\
 -H "accept-encoding: gzip"\
`


mrin9 (approved):

mrin9:

Issue #1032, opened 2024-09-25
improve tree shakability - reduce size 


christophe-g:
First of all - thanks million for this. It is great, and I love it is using lit ! 

It would be great if we could make better use of ESM and avoid potential code dupe when consuming Rapidoc. 

For instance, I would like to avoid re-importing `lit`, `marked` or `Prism` (contained in the dist file) if I consume it in a project that already import those dependencies.

christophe-g:
Related discussion: https://github.com/shoelace-style/shoelace/issues/559

mrin9:
please refer my comments in the PR #1033 and the reason for closing it. You may create a new issue as a feature-request to generate `ESM` build if you like to track it and may contribute towards it when we start that branch

PR #1033, opened 2024-09-25
refactor: standard ESM import for all local imports

christophe-g:
One step forward towards #1032 ...

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,6 +24,7 @@
     "globalThis": false // means it is not writeable
   },
   "rules": {
+    "import/extensions": "ignorePackages",
     "max-len": [1, 300, 2, {"ignoreComments": true}],
     "no-debugger": 0,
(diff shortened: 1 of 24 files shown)


mrin9:
@christophe-g thanks for the PR. Unfortunately I wont be able to accept this for the below reasons

- The PR do not perform anything other than removing a [`resolve-alia`](https://webpack.js.org/configuration/resolve/#resolvealias) which we do like to maintain be it ESM or no ESM. having these alias do not have any effect on the ability to generate ESM module. Instead what matters is there shouldn't be any `require` syntax
- We do have plans to generate ESM builds soon (likely the work will start by end of oct)
- We will be moving towards using Vite instead of Webpack as our bundling tool

christophe-g:
> We will be moving towards using Vite instead of Webpack as our bundling tool

That's excellent, thanks !

Issue #1035, opened 2024-10-01
Wrong request ouptut with 204 No content


Zstorm999:
I noticed a bug that I think has been introduced recently (I did not encounter it before).

I have a request that returns a 204 No content, and so has no `content-type` header. 

It seems that this generates an error at this line

https://github.com/rapi-doc/RapiDoc/blob/0825d00aa8e96b9fc3ceb1bc42594a337df2337c/src/components/api-request.js#L1531

and is incorrectly caught here

https://github.com/rapi-doc/RapiDoc/blob/0825d00aa8e96b9fc3ceb1bc42594a337df2337c/src/components/api-request.js#L1620-L1630

The overall effect is that the interface does not display the output of the request (it did show an empty block before), and shows an error instead:

![image](https://github.com/user-attachments/assets/1bf69436-96cf-4b8a-a975-1e1a1ecd49cb)

A simple fix would be to check for the http code 204, and skip processing content if it is present.

PR #1036, opened 2024-10-01
fix: do not parse header content-type if empty

Zstorm999:
This is a proposal fix for #1035. It avoids the error by parsing the `content-type` header only if its value is not `null`


mrin9 (approved):

mrin9:
thanks for your contribution !!!

PR #1064, opened 2024-12-16
fix/improve deduplication of header values

### LINKED ISSUE #1063: Header examples with comma are displayed incorrectly in curl command
When a header is included in the openapi spec and its example contains a comma, the header example is displayed incorrectly in the example curl command. This is especially an issue for date-time headers:
```yaml
      - name: If-Modified-Since
        in: header
        schema:
          format: date-time
          examples:
          - "Wed, 21 Oct 2015 07:28:00 GMT"
```
is displayed as:
image

this was probably introduced with #1024

derklaro:
Fixes the header deduplication for curl requests introduced in #1024

When the old code receives the following input:
`"hello, world, some, cool, World, words, IN, a,ROW,TesT,Hello,testing"` 
as the header value, it deduplicates the it to
`hello, ROW, TesT, Hello, testing`

Which has 3 problems:
1. the header values are not deduplicated correctly, resulting in the fact that non-duplicate elements are removed as well
2. the spacing around the comma is not presevered and every element is exactly one space apart after deduplicating
3. duplicate elements with different casing are left in the resulting header value

When the new code receives the same input, it deduplicates them correctly, while fixing the other 2 mentioned issues as well:
`hello, world, some, cool, words, IN, a,ROW,TesT,testing`

Closes #1063

### DIFF
--- a/src/components/api-request.js
+++ b/src/components/api-request.js
@@ -1937,9 +1937,21 @@ ${responseContent}</pre
     curl = `curl -X ${this.method.toUpperCase()} "${curlUrl}" \\\n`;
 
     fetchHeaders.forEach((value, key) => {
-      let tempHeaderArray = value.split(',');
-      tempHeaderArray = tempHeaderArray.map((el) => el.trim()).filter((string, index) => tempHeaderArray.indexOf(string) === index);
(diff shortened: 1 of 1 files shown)


mrin9:
thank you for your contribution !!!

PR #1070, opened 2025-01-07
feat: allow ctrl+click enum array values to add/remove

John-Paul-R:
Quick and dirty feature-add allowing quickly toggling enum values in and out of an array field.

https://github.com/user-attachments/assets/7a6c434d-8d71-418a-a134-e74b86d254c4

May explore a variant that has no layout shift in the future, because the current reflow is a bit awkward imo.


mrin9:
thanks will take a look soon

PR #1117, opened 2025-08-21
fix: Do not show braces when the child is a OneOf

BentEngbers:
This minor change to how the parent of a `ONE OF` element is rendered makes the number of braces match the json format (see example in the comment below)

Checks from the contributing guide:

- [x] yarn build does not emit any errors 
- [x] the `./dist` folder is empty


BentEngbers:
# Example
Consider the following spec.yaml:
```yaml
openapi: 3.1.0
info:
  title: Simple API
  version: 1.0.0
  description: API with various nullable property examples
paths:
  /items:
    post:
      operationId: test
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                # Polymorphism and Null
                payment_method:
                  description: Can be a credit card, bank account, or null if no payment method is on file.
                  oneOf:
                    - type: object
                      properties:
                        card_number:
                          type: string
                    - type: object
                      properties:
                        bank_account:
                          type: string
                    - type: "null"
                # Object with Nullable Keys
                address:
                  description: An object representing an address where some fields can be explicitly null.
                  type:
                    - object
                    - "null"
                  properties:
                    street:
                      type: string
                    city:
                      type: string
                    postal_code:
                      type: string
                    country:
                      type: string
      responses:
        "200":
          description: OK
```

This used to render like this: 
image

Note that for `payment_method`, the current schema could be interpreted as invalid json:
```json
  "payment_method": {{
    "card_number": "1234"
  }}
```
The intended structure is:
```json
  "payment_method": {
    "card_number": "1234"
  }
```
After this change, the extra braces are removed (for example, the `{` after the key `payment_method`):
image

iljamaas:
What could I do to help this one getting merged?

